PR #399, opened 2026-05-05
feat(web): mid-session item reflection sheet (Pencil hZfKR)

jonyardley:
## Summary

Closes the gap flagged in the [interstitial reflection memory](https://github.com/jonyardley/intrada/blob/main/docs/roadmap.md): mid-session scoring UX. Pencil reference: `hZfKR` ("Practice / Item Transition" interstitial).

When the user taps **Next Item** or **Finish Session**, a bottom sheet now rises to capture **self-rating + achieved tempo + notes** for the item that just completed. The dispatch is deferred until the user hits **Continue** inside the sheet (sends `UpdateEntryScore` + `UpdateEntryTempo` + `UpdateEntryNotes` for the entry id, then `NextItem` / `FinishSession`), or fires immediately on **Skip scoring** without recording reflection data.

The summary screen still has the same scoring controls, so users who prefer one big batch at the end get that workflow; this just lets practice-as-you-go users capture data while it's fresh.

All underlying core events (`UpdateEntryScore`/`Tempo`/`Notes`) and entry fields (`score`, `achieved_tempo`, `notes`) already existed for the summary phase — **no FFI / DB / model changes**.

## UX decisions

- **Sheet opens for the last item too**, not just between items — user can reflect on the final item before hitting the summary.
- **Sheet does NOT auto-open** when planned duration elapses (would feel jarring while the user is still mid-phrase). Tap-only.
- **Skip scoring** is the escape hatch — text link below Continue.
- **Tempo input** uses the same `validate_achieved_tempo_input()` rule as the summary screen so the two surfaces share validation.
- Pre-populates from the entry's existing `score`/`achieved_tempo`/`notes` (so a revisit works), reset each time a fresh target arrives.

## Implementation

- `<ItemReflectionSheet>` wraps `<BottomSheet>` and exposes a snapshot-based API: parent passes an `ItemReflectionTarget` (entry id + initial values) at open time so the sheet doesn't race with view-model advancement.
- `dispatch_advance` is a `Callback<bool>` (capture vs skip) so both Continue and Skip can call it — a plain `move` closure with this many captures is not Copy and would be moved into whichever handler ran first.
- The Next/Finish CTA in `<SessionTimer>` now just opens the sheet; the `on_advance` callback re-reads the view model at advance time to decide between `NextItem` and `FinishSession` (dispatch always matches the actual session state, not the state when the sheet opened).

## Test plan

- [x] `cargo fmt --all -- --check`
- [x] `cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings`
- [x] `cargo test -p intrada-core` — 247 passed
- [ ] Manual: tap Next Item mid-session → sheet opens with "Up Next: [next title]", rate 4, set tempo, hit Continue → next item starts and entry is scored
- [ ] Manual: tap Skip scoring → advances without recording
- [ ] Manual: tap Finish Session on last item → sheet shows "Last Item" header → Continue routes to summary with entry scored
- [ ] Manual: re-open sheet after a partial reflection → previous values pre-populated

⚠️ **Pre-existing e2e infra failure**: the suite is currently broken on `main` too (`Failed to load routines: JSON serialization error` on every test). Not introduced by this PR; tracking separately.


jonyardley:
## Update: bug found + fixed during visual verification

Pushed [`f9f0786...`](https://github.com/jonyardley/intrada/commit/abebeac432e31ef8a76834e59b186a920596d51c) — the mid-session reflection sheet's data wasn't actually persisting. Caught it on a smoke run through the preview server: rated Moonlight Sonata 4 + tempo 120 + notes "Felt solid" via the sheet, advanced, finished with item 2 — summary screen showed both rows with empty score/tempo/notes.

Root cause: `UpdateEntryScore` / `UpdateEntryTempo` / `UpdateEntryNotes` were gated on `SessionStatus::Summary` only. The sheet fires them while the session is still `Active` → silent no-op.

Fix:
- Core `entry_for_update_mut(model, entry_id)` helper that finds an entry in either Active or Summary phase. The `entry.status == Completed` invariant stays as the real safety check.
- Flipped dispatch order in the sheet: advance (`NextItem` / `FinishSession`) FIRST, then dispatch the per-entry update events. `NextItem` is what flips the entry to `Completed` status; dispatching update-then-advance would fail the Completed check even with the relaxed phase gate.

Tests:
- Renamed `test_update_entry_score_only_works_during_summary` → `test_update_entry_score_rejected_on_pending_entry` (the assertion is now precise: Pending entries can't be scored regardless of phase).
- Added 3 regressions for the mid-session path. **250 pass** (was 247).

Verified manually: 2-item session → rate 4 + tempo 120 mid-session → rate 5 + tempo 92 on finish → summary now shows **Avg 4.5/5** with both items' data populated. ✅

jonyardley:
## Self-review fixups pushed

Ran the `superpowers:code-reviewer` agent against this PR — surfaced 3 must-do items + a few suggestions. Just pushed [`a57202d`](https://github.com/jonyardley/intrada/commit/a57202d65e7fcdf2437b4e20360c9d353836af8b) addressing the must-dos:

**1. Last-item regression test.** Both the non-last (Active) and last-item (Summary, post-`FinishSession`) paths now resolve via `entry_for_update_mut`, but only the non-last path was tested. Added `test_update_entry_score_works_on_last_item_after_finish_session`.

**2. Edge-case tests.**
- `test_update_entry_score_unknown_entry_id_is_silent_noop` — pins that a stale entry id (from session recovery / new session between sheet-open and Continue) is a silent no-op for Score + Tempo.
- `test_update_entry_score_rejected_on_skipped_entry_in_active_phase` — `SkipItem` produces `EntryStatus::Skipped`; the status gate should reject scoring it mid-session.
- **253 pass** (was 250).

**3. Backdrop-dismiss semantics documented.** `BottomSheet`'s `on_close` (backdrop / swipe / Escape) bypasses both `on_continue` and `on_skip` — dismissal closes without advancing, no draft preservation. That's the right behaviour (matches iOS Mail-compose) but was invisible in the component doc. Spelled out now.

**4. Design catalogue entry** for `<ItemReflectionSheet>` per CLAUDE.md ("New components get a showcase entry"). Section + TOC link + working demo pre-seeded against a sample target.

Drive-by tightening: dispatch-order rationale comment in `dispatch_advance` (the original referenced the now-removed Summary-only gate), and a comment in `on_advance_tap` pinning the target-set-before-open ordering invariant.

**Follow-up flagged** for a separate PR ([Suggestion] from the review): extract a `<RatingChips>` primitive — the `1u8..=5` chip markup is now duplicated between summary screen and reflection sheet, which is the second-surface trigger for CLAUDE.md's "extend, don't clone" rule.

PR #413, opened 2026-05-06
feat(web): re-introduce Save as Set in session review sheet

jonyardley:
## Summary
- Mounts the existing `<SetSaveForm>` at the bottom of the `SessionReviewSheet` body, gated on a non-empty building setlist via a shared `has_entries` signal so the sheet's empty state stays clean.
- `on_save` dispatches `Event::Set(SetEvent::SaveBuildingAsSet { name })` — the core handler is already implemented and tested.
- Re-enables the previously-skipped `save set from session builder` e2e test.

Closes the **first half** of [intrada#390](https://github.com/jonyardley/intrada/issues/390). The second half (Load Set into the builder with a merge/replace prompt) is a follow-up PR.

The component itself was already in the codebase (and still mounted in the post-session summary as `SaveSummaryAsSet`) — this PR is the wiring that brings it back to the builder side after the strip-back.

## Test plan
- [x] `cargo fmt --check && cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings`
- [x] `cargo test -p intrada-core` (253 pass)
- [x] Empty-state visual check in preview: review sheet opens, no Save form rendered when setlist is empty
- [ ] CI runs the un-skipped Playwright test end-to-end
- [ ] Manual: open builder, add an item, tap Review session, see "Save as Set" pill at bottom; expand, name it, Save → set appears in `/routines`

### DIFF
--- a/crates/intrada-web/src/components/session_review_sheet.rs
+++ b/crates/intrada-web/src/components/session_review_sheet.rs
@@ -1,20 +1,20 @@
 use leptos::prelude::*;
 
-use intrada_core::{Event, SessionEvent, ViewModel};
+use intrada_core::{Event, SessionEvent, SetEvent, ViewModel};
 
-use crate::components::{BottomSheet, EditorEntry, EntryListEditor};
+use crate::components::{BottomSheet, EditorEntry, EntryListEditor, SetSaveForm};
 use intrada_web::core_bridge::{process_effects, process_effects_with_core};
 use intrada_web::types::{IsLoading, IsSubmitting, SharedCore};
 
 /// Bottom sheet that opens from the Review session CTA in the builder.
-/// Stripped to the essentials per the Pencil `AYx23` design: session
-/// intention, drag-reorderable entry list with a per-row remove button,
-/// and a total. The Start action lives in the sheet's nav bar (right side,
-/// opposite Cancel) — iOS Mail-compose pattern.
+/// Per Pencil `AYx23`: session intention, drag-reorderable entry list
+/// with a per-row remove button, total, and a "Save as Set" form at the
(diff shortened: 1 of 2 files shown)


jonyardley:
## Self-review

Ran `superpowers:code-reviewer` against the staged diff before pushing.

### BLOCKERS
None.

1. **Reuse the existing entries-empty predicate.** The reviewer flagged that the parent `SessionReviewSheet` already computes `building_setlist.entries.is_empty()` for `nav_action_disabled`, and the new `<Show>` predicate duplicates that logic in the body component. Refactored to a shared `has_entries: Signal<bool>` at the top of `ReviewSheetBody` so the gate matches the nav-action signal exactly. Markup uses `move || has_entries.get()`.
2. **e2e brittleness.** Going to `/routines` after save will break when the URL migration drops the route. Added `// TODO(#390 pt 2): switch to /library once the URL migration drops /routines …` so the next person knows why it's odd.

### NITS
- Reactivity in `Show when=` is correctly tracked (`view_model.with(|vm| …)` registers the read).
- `core_save_set.clone()` capture mirrors the precedent in `session_summary.rs:314`; cheap clone on `SharedCore` (`Rc<RefCell<…>>`-shaped), no leak.
- Doc comment now accurately reflects that save-as-set is back in the sheet; `#389` reference for per-entry controls preserved.
- No design-system violations — `SetSaveForm` is the existing primitive, just being re-mounted.

### Notes
- Empty-state visually verified in preview (Show gate hides the form).
- Populated state not visually verified locally (dev API is 500'ing); CI's Playwright run will exercise the full save flow end-to-end via the un-skipped test.

PR #414, opened 2026-05-07
feat(web): re-introduce Load Set affordance in session builder

jonyardley:
## Summary
- Mounts the existing `<SetLoader>` at the top of the `SetlistBuilder`, gated on the existing `setlist_empty` signal so it disappears once the user adds an item.
- Refactors `SetLoader` from raw markup (`rounded-lg bg-surface-secondary px-3 py-2 …`) to `<AccentRow bar=AccentBar::Teal>` + `<Button variant=Secondary>` — mirrors `LibrarySetCard`'s chrome so saved Sets read consistently across Library and the builder.
- Re-exports the component (was deliberately unexported during the strip-back).
- Re-enables the previously-skipped `load set into session builder` e2e test, with a tightened listitem-scoped selector.

Closes [intrada#390](https://github.com/jonyardley/intrada/issues/390) (the second half — the first half landed in [intrada#413](https://github.com/jonyardley/intrada/pull/413)).

### Why no merge/replace prompt?
The issue calls out merge/replace as TBD. Empty-only gating sidesteps the question entirely: `LoadSetIntoSetlist` already appends, and append-into-empty == replace-from-zero. Avoids a Crux core change for behaviour the user may never need (their flow at this point is "load a set OR pick items individually", not "load on top of items I already picked"). If we discover users do want this, we'd add `SetEvent::ReplaceSetlistWithSet` and an inline merge/replace choice — easy follow-up.

## Test plan
- [x] `cargo fmt --check && cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings`
- [x] `cargo test -p intrada-core` (253 pass)
- [x] Empty-state visual check: builder renders, no "Saved Sets" panel when no sets exist (local API was 500'ing so positive state not exercised locally)
- [ ] CI runs the un-skipped Playwright test end-to-end with mockApi seed
- [ ] Manual: visit `/sessions/new` → Custom Session, see "Saved Sets" panel above search bar; tap Load on a set → entries appear in the library list as selected, panel hides, Review session enabled

## Follow-ups (not in this PR)
- If saved-sets count grows large the loader needs `max-h` / scroll. Surface as an issue if it becomes a real problem.

### DIFF
--- a/crates/intrada-web/src/components/mod.rs
+++ b/crates/intrada-web/src/components/mod.rs
@@ -84,12 +84,11 @@ pub use page_heading::PageHeading;
 pub use progress_ring::ProgressRing;
 pub use pull_to_refresh::PullToRefresh;
 pub use rating_chips::RatingChips;
-// SetLoader is currently unused — kept around for the planned set
-// flows revisit (see PR conversation). Re-export when something needs it.
 pub use section_label::SectionLabel;
 pub use session_review_sheet::SessionReviewSheet;
 pub use session_summary::SessionSummary;
 pub use session_timer::SessionTimer;
+pub use set_loader::SetLoader;
 pub use set_save_form::SetSaveForm;
 pub use setlist_builder::SetlistBuilder;
 pub use setlist_entry::SetlistEntryRow;
--- a/crates/intrada-web/src/components/set_loader.rs
+++ b/crates/intrada-web/src/components/set_loader.rs
@@ -2,14 +2,21 @@ use leptos::prelude::*;
 
 use intrada_core::{Event, SetEvent, ViewModel};
 
-use crate::components::Card;
+use crate::components::{AccentBar, AccentRow, Button, ButtonVariant};
 use intrada_web::core_bridge::process_effects;
 use intrada_web::types::{IsLoading, IsSubmitting, SharedCore};
 
-/// Shows saved sets and lets the user load one into the current setlist.
+/// Lists the user's saved Sets and lets them load one into the current
+/// (empty) building setlist. Renders nothing when there are no saved Sets.
 ///
-/// Each set is displayed as a row with name, entry count, and a Load button.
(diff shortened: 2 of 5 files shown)


jonyardley:
## Self-review

### BLOCKERS
None.

1. **e2e selector ambiguity** — \`getByRole("button", { name: "Load" }).click()\` would collide if the seed ever had >1 set. Tightened to \`getByRole("listitem").filter({ hasText: "Morning Warm-up" }).getByRole("button", { name: "Load" })\` so it stays unambiguous.

### SUGGESTIONS (deferred — see PR body / follow-ups)
2. **Many-saved-sets edge case** — no virtualization or \`max-h\`/scroll on the \`<ul>\`. With ~50+ saved sets the loader pushes the library list off-screen on mobile. Acceptable to defer at current scale; called out in the PR body as a follow-up.

- **Empty-only gating** — reviewer agreed this is the right call: \`LoadSetIntoSetlist\` already appends, append-into-empty == replace-from-zero, no core change needed.
- **Reactivity** — \`setlist_empty\` is a \`Signal::derive\` (Send + Sync in Leptos 0.8); reads inside \`Show when=\` and \`Callback::new\` body are fine.
- **AccentBar::Teal** — consistent with \`LibrarySetCard\`. Right call.
- **Design system** — refactor cleanly removes the raw \`rounded-lg bg-surface-secondary …\` and the bespoke \`<button class="text-xs …">\`. No spacing-token violations.

### NITS
- \`count_label\` builds a \`String\` per row unconditionally — fine, not worth optimizing.
- Catalogue entry text updated to reflect the new primitives; no separate live render exists for SetLoader, which is fine.

### Notes
- Empty-state visually verified in preview (panel correctly absent when no sets).
- Positive state (panel visible with saved sets) not exercised locally — dev API 500'ing — but the un-skipped Playwright test exercises it end-to-end with mockApi seed.

PR #415, opened 2026-05-07
refactor(web): drop /routines surface, fold Sets into Library

jonyardley:
## Summary
Final IA-shift step. Sets now live entirely under the Library Sets type-tab; the legacy \`/routines\` listing surface retires.

- **Bottom nav**: dropped from 4 tabs to 3 (Library / Practice / Analytics). Sets-icon SVGs removed from \`bottom_tab_bar.rs\`.
- **Desktop header**: Sets link removed.
- **\`?type=\` query support** on \`LibraryListView\` so deep links and the redirect can preselect a tab. Reads with \`with_untracked\` — URL is an entry point, not a live source of truth.
- **\`/routines\` redirect**: \`<Redirect path="/?type=set" options=replace:true />\` so legacy bookmarks land on the right tab and the dead URL doesn't pollute the back stack.
- **Deleted** \`views/sets.rs\` (\`SetsListView\`). \`LibrarySetCard\` on the Library Sets tab is the sole render surface now.
- **\`set_edit.rs\`** back-links and post-save navigation retargeted to \`/?type=set\`.
- **e2e**: \`sets.spec.ts\` rewritten — empty-state and pre-seeded-set tests now land on \`/?type=set\`, assert \`aria-selected\` on the Sets tab, check \`LibrarySetCard\`'s \`/library/sets/{id}\` href and \`"N items"\` count format. New test asserts \`/routines\` → \`/?type=set\` redirect.

\`/routines/:id/edit\` is **intentionally kept** — that URL's migration is a separate future PR (changes the Set Detail Edit link and the LibrarySetCard context-menu Edit). Out of scope to keep the diff focused.

Closes the IA-shift sequence. Stacks on [intrada#414](https://github.com/jonyardley/intrada/pull/414).

## Trade-off
Sets goes from a top-level nav destination to a sub-tab inside Library. In-app discoverability narrows; screen-reader users navigating by landmarks lose one stop. Mitigations: the \`/routines\` redirect preserves external deep links, and the Library Sets tab is reachable in 1 tap from \`/\`. Worth a roadmap note if we observe drop-off; flagging here so it's not a silent change.

## Test plan
- [x] \`cargo fmt --check && cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings\`
- [x] \`cargo test -p intrada-core\` (253 pass)
- [x] Manual in preview: \`/routines\` → \`/?type=set\`, Library heading visible, Sets tab \`aria-selected="true"\`, "No saved sets yet" empty-state shown
- [x] Manual in preview: bottom nav now has 3 entries (Library / Practice / Analytics)
- [ ] CI runs the rewritten Playwright tests end-to-end

## Follow-ups (not in this PR)
- Migrate \`/routines/:id/edit\` → \`/library/sets/:id/edit\` (touches Set Detail Edit link + LibrarySetCard context-menu Edit)


jonyardley:
## Self-review

### BLOCKERS
None.

1. **Cleaner redirect.** Reviewer flagged my Effect-based \`RoutinesRedirect\` as slightly heavier than needed — there's a single-frame paint where the route mounts before the navigate runs. Swapped to \`<Redirect path="/?type=set" options=replace:true />\` from \`leptos_router::components\` (still present in 0.8.13). SSR-safe and idiomatic; deleted the custom component entirely. Verified the redirect still works in preview.
2. **Document the \`with_untracked\` choice.** Added a comment block in \`library_list.rs\` explaining that the URL is intentionally treated as an entry point rather than a live source of truth (so a future maintainer doesn't "fix" it by switching to \`with\` and breaking tab clicks).

### SUGGESTIONS (deferred — see PR body)
3. **\`set_edit\` BackLink could deep-link to \`/library/sets/{id}\`** instead of the Sets list, so iPad split-view users land back on the just-edited Set Detail. Out of scope here — flagged for the upcoming \`/routines/:id/edit\` migration PR.
4. **Accessibility / discoverability trade-off** of dropping a top-level nav target. Acknowledged in the PR body.

- **\`Effect::new\` semantics** — reviewer agreed no infinite-loop risk in the original form; \`<Redirect>\` is just cleaner.
- **\`filter_from_query\` parsing** — accepts singular/plural for free; covers all reasonable inputs.
- **No straggler \`/routines\` listing links** — reviewer scanned and confirmed; \`library_set_card.rs\` / \`set_edit.rs\` / \`set_detail.rs\` references all point at \`/routines/:id/edit\` which is intentionally retained.
- **\`aria-selected\` assertion** is solid — \`LibraryFilterTabs\` sets it from \`active_filter\` (verified at \`library_filter_tabs.rs\`).
- **\`SetsListView\` deletion** is clean — no other re-exports or imports.

### Notes
- Manually verified in preview: \`/routines\` → \`/?type=set\`, Library heading visible, Sets tab \`aria-selected="true"\`, mobile bottom nav shows 3 tabs only.
- Stacks on #414 — base for this PR is \`feat/load-set-in-builder\`. Will rebase onto main once #414 merges.

Issue #309, opened 2026-04-25
ios: background audio Swift plugin for lock-screen practice timers


jonyardley:
## Summary

Custom Tauri Swift plugin to keep practice session timers running when the screen locks. Without it, the timer stops when the phone times out — broken product experience for real practice sessions.

## Why deferred

Not blocking during active development (screen can stay on). Becomes critical before public TestFlight/App Store submission. File now so it's tracked.

## Scope

- Tauri plugin under `crates/intrada-mobile/plugins/background-audio/`
- `AVAudioSession.sharedInstance().setCategory(.playback)` + `setActive(true)` on session start
- `UIBackgroundModes: [audio]` in Info.plist
- `MPNowPlayingInfoCenter` for lock-screen Now Playing (pairs with Media Session API in JS)
- JS surface: `invoke('plugin:background-audio|begin_session')` / `end_session()` / `set_now_playing({title})`
- Tauri command IPC bridge

## Spec reference

`specs/tauri-leptos-ios-shell.md` §8 P0 (background audio session, parity-blocker).

PR #431, opened 2026-05-07
feat(web): wall-clock timer derive (Phase A of #309)

jonyardley:
## Summary
**Phase A** of [intrada#309](https://github.com/jonyardley/intrada/issues/309) per [specs/background-audio-plugin.md](https://github.com/jonyardley/intrada/blob/main/specs/background-audio-plugin.md). Replaces the JS \`setInterval\` counter in \`SessionTimer\` with a wall-clock derive: \`elapsed = Utc::now() - current_item_started_at\`. The 1Hz interval still fires; it just triggers a recompute now rather than incrementing.

**Why it matters even before iOS work**: Chrome throttles \`setInterval\` on backgrounded tabs to 1Hz, so the old tick-counter silently under-reported elapsed time when a user practiced with the tab in the background. The wall-clock derive catches up to truth in one tick on resume. Same logic also lays the foundation for Phase C (Tauri WKWebView suspension on iOS lock-screen).

## Changes

1. **\`crates/intrada-core/src/model.rs\`** — \`ActiveSessionView\` gains \`current_item_started_at: String\` (RFC3339 UTC), populated from the existing \`ActiveSession.current_item_started_at\`. Serde-additive field; the only consumer is the WASM shell in the same workspace.
2. **\`crates/intrada-web/src/components/session_timer.rs\`** — \`recompute()\` parses the anchor and computes elapsed seconds. Initial seed (covers crash-recovery rehydration). Effect re-derives on \`current_item_started_at\` change so item-advance doesn't briefly show stale elapsed before the next tick. Three explicit \`elapsed_secs.set(0)\` reset calls dropped — derive self-corrects.

## Tier note
Per CLAUDE.md, FFI / ViewModel changes attract a tier-up. This is the additive serde-defaulted field case (both ends of the bridge compile from the same workspace; no breakage risk). Treating as Tier 2 with the override flagged. Self-reviewer agreed.

## Verification
- \`cargo fmt --check && cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings\` clean
- \`cargo test -p intrada-core\` 253 pass
- **Manual in preview** with API + web running locally:
  - Started a Custom Session → timer ticks 00:01, 00:02, …
  - Measured wall-clock: 17 real seconds elapsed → timer delta 17 (matches reality, not interval-fire count)
  - Simulated backgrounding via \`document.visibilityState = 'hidden'\` + 10s sleep + back to \`'visible'\` → timer correctly tracked all elapsed seconds during the hidden window
  - No console errors

## Crash-recovery side benefit
A session rehydrated 7 minutes in now shows \`07:00\` immediately rather than \`00:00\` climbing up. Same code path; happens to be correct now.

## Phasing
- **A (this PR)** — wall-clock derive; web-compatible foundation
- B — Tauri plugin scaffold (Rust + Swift, commands stubbed)
- C — Swift impl: AVAudioSession + MPNowPlayingInfoCenter + interruption handling
- D — Sentry instrumentation, error surfacing

## Test plan
- [x] fmt + clippy + core tests
- [x] Manual visibility-change recovery test in preview
- [ ] CI runs e2e (active-session test should still pass — same RwSignal\<u32\> elapsed_secs)
- [ ] Manual on physical iOS device once a TestFlight build exists (Phase C)

### DIFF
--- a/crates/intrada-core/src/model.rs
+++ b/crates/intrada-core/src/model.rs
@@ -225,6 +225,12 @@ pub struct ActiveSessionView {
     pub current_position: usize,
     pub total_items: usize,
     pub started_at: String,
+    /// Wall-clock anchor (RFC3339 UTC) for the *current item*. Resets to
+    /// "now" on each item advance (Next / Skip). The shell derives the
+    /// per-item elapsed timer from `Utc::now() - current_item_started_at`
+    /// rather than incrementing a counter — survives WebView suspension /
+    /// tab backgrounding without drift.
+    pub current_item_started_at: String,
     pub entries: Vec<SetlistEntryView>,
     pub session_intention: Option<String>,
     pub current_rep_target: Option<u8>,
@@ -298,6 +304,7 @@ pub fn build_active_session_view(active: &ActiveSession) -> ActiveSessionView {
         current_position: active.current_index,
         total_items: active.entries.len(),
         started_at: active.session_started_at.to_rfc3339(),
+        current_item_started_at: active.current_item_started_at.to_rfc3339(),
         entries: active.entries.iter().map(entry_to_view).collect(),
         session_intention: active.session_intention.clone(),
         current_rep_target: current.rep_target,
--- a/crates/intrada-web/src/components/session_timer.rs
+++ b/crates/intrada-web/src/components/session_timer.rs
@@ -1,3 +1,4 @@
+use chrono::DateTime;
 use leptos::prelude::*;
 use wasm_bindgen::closure::Closure;
 use wasm_bindgen::JsCast;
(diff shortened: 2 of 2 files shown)


jonyardley:
## Self-review

### BLOCKERS
None.

1. **Equality guard documentation** — the \`if elapsed_secs.get_untracked() != next\` check skips redundant signal sets when the second hasn't ticked over. Reviewer flagged that future readers might assume it defends against re-entrancy. Added a comment block explaining it's an intentional optimization to avoid up-to-N renders per second on sub-second 1Hz fire after a recompute trigger.
2. **Visibility test** — reviewer suggested 30 seconds of literal background/foreground testing as cheap insurance. Done: \`document.visibilityState = 'hidden'\` + 10s sleep + restore → timer correctly tracked all elapsed seconds. (Chrome's actual setInterval throttling on a hidden tab can't be perfectly simulated via DevTools alone, but the algebra is sound regardless: the next interval fire pulls wall-clock truth.)

- **Effect ordering** — Leptos 0.8 \`Effect::new\` runs after the signal that triggered it has updated, so reads inside see the new \`current_item_started_at\`. Crux bridge sets the VM in a single transaction per process_effects call. No staleness risk.
- **\`Closure::<dyn Fn()>::new(recompute)\`** — \`recompute\` is \`Fn + 'static\` because all captures are \`Copy\` signal handles. wasm32 single-threaded so no Send concern.
- **\`with_timezone(&Utc)\`** is needed (\`parse_from_rfc3339\` returns \`DateTime<FixedOffset>\`; subtraction with \`Utc::now()\` requires same TZ type). Reviewer suggested it was redundant — checked; it isn't.

### Notes
- Tier-2-with-override decision flagged in PR body per CLAUDE.md.
- No deferred items from this PR.

PR #441, opened 2026-05-07
feat(mobile): scaffold background-audio Tauri plugin (Phase B of #309)

jonyardley:
## Summary
**Phase B** of [intrada#309](https://github.com/jonyardley/intrada/issues/309) per [\`specs/background-audio-plugin.md\`](https://github.com/jonyardley/intrada/blob/main/specs/background-audio-plugin.md). Phase A (#431, wall-clock timer) merged. This PR scaffolds the local Tauri plugin so the IPC channel is proven end-to-end with no native side-effects yet. Phase C swaps in the Swift \`AVAudioSession\` + \`MPNowPlayingInfoCenter\` impl behind the same command names.

## What's in
- **\`crates/intrada-mobile/plugins/background-audio/\`** — new local plugin crate, added to the workspace. Three \`#[tauri::command]\` async fns (\`begin_session\`, \`set_now_playing\`, \`end_session\`) all return \`Ok\` with no native side-effects. \`tauri-plugin\` build script generates the permission TOML / schema / reference.md from \`build.rs\`.
- **\`src-tauri/Cargo.toml\` + \`lib.rs\`** wired so the plugin loads alongside haptics + deep-link.
- **\`capabilities/mobile.json\`** adds \`background-audio:default\` so the WebView is permitted to invoke the commands at runtime.
- **\`intrada-web/src/background_audio.rs\`** — \`wasm_bindgen\` inline-JS bindings matching the haptics pattern. try/catch around \`window.__TAURI__.core.invoke\`; on web / E2E browsers the global is absent and the bindings no-op silently.
- **\`session_timer.rs\`** — new Effect tracks \`current_item_started_at\` across renders to detect session-start (None→Some / \`begin_session\`), item-advance (anchor change / \`set_now_playing\`), session-end (Some→None / \`end_session\`) with a single closure + \`prev_anchor\` diff.

## Tier
Per CLAUDE.md, Tauri plugin IPC contract changes are Tier 3 territory. Phase B is scaffold-only (no Swift bridge, no native effects, commands are Rust stubs that return Ok); treating as Tier 2 with the override flagged. Phase C will be the genuinely Tier 3 PR.

## Verification
- \`cargo fmt --check && cargo clippy -p intrada-web --target wasm32-unknown-unknown --no-default-features -- -D warnings\` ✅
- \`cargo clippy -p tauri-plugin-background-audio -- -D warnings\` ✅
- \`cargo test -p intrada-core\` 253 pass ✅
- \`cargo build -p intrada-mobile\` ✅ (plugin loads, mobile shell compiles)
- Manual web in preview: started a session → no console errors, timer ticks correctly, plugin bindings invoke harmlessly (catch swallows the missing \`__TAURI__\` global)

## Test plan (deferred)
- [ ] CI: e2e tests still pass (call sites are no-ops on web; nothing observable to add)
- [ ] Phase C will need physical-device testing per CLAUDE.md "test on device" rule for Tauri WebView interactions

## Phasing
- A — wall-clock timer (#431, **merged**)
- **B — plugin scaffold (this PR)**
- C — Swift impl: AVAudioSession + MPNowPlayingInfoCenter + interruption handling
- D — Sentry instrumentation, error surfacing, cleanup audit

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,11 +67,12 @@ jobs:
       - uses: taiki-e/install-action@v2
         with:
           tool: cargo-nextest
-      # intrada-mobile (Tauri 2) requires GTK/glib system libs not present on
-      # Ubuntu runners — it's an iOS-only host with no meaningful unit tests.
-      - run: cargo nextest run --workspace --exclude intrada-mobile
+      # intrada-mobile + its plugins (Tauri 2) require GTK/glib system libs
+      # not present on Ubuntu runners — iOS-only crates with no meaningful
+      # unit tests on Linux.
+      - run: cargo nextest run --workspace --exclude intrada-mobile --exclude tauri-plugin-background-audio
       - name: Doc tests
-        run: cargo test --doc --workspace --exclude intrada-mobile
+        run: cargo test --doc --workspace --exclude intrada-mobile --exclude tauri-plugin-background-audio
 
   clippy:
     name: Clippy
@@ -87,8 +88,8 @@ jobs:
         with:
           prefix-key: "native"
           save-if: "false"
-      # intrada-mobile excluded: Tauri 2 needs GTK/glib system libs on Linux.
-      - run: cargo clippy --workspace --exclude intrada-mobile -- -D warnings
+      # intrada-mobile + its plugins excluded: Tauri 2 needs GTK/glib system libs on Linux.
+      - run: cargo clippy --workspace --exclude intrada-mobile --exclude tauri-plugin-background-audio -- -D warnings
       - name: Clippy (WASM target)
         run: cargo clippy --target wasm32-unknown-unknown -p intrada-web -- -D warnings
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3035,6 +3035,7 @@ dependencies = [
  "serde_json",
  "tauri",
  "tauri-build",
+ "tauri-plugin-background-audio",
  "tauri-plugin-deep-link",
  "tauri-plugin-haptics",
 ]
@@ -6816,6 +6817,16 @@ dependencies = [
  "tauri-utils",
  "toml 0.9.12+spec-1.1.0",
  "walkdir",
+]
+
+[[package]]
+name = "tauri-plugin-background-audio"
+version = "0.0.1"
+dependencies = [
+ "serde",
+ "tauri",
+ "tauri-plugin",
+ "thiserror 2.0.18",
 ]
 
 [[package]]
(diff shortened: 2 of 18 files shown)


jonyardley:
## Self-review

Ran \`superpowers:code-reviewer\` against the diff before pushing. No blockers; applied three fixes before opening.

### Applied (review fixes commit)
1. **Dropped \`Manager\` import + \`_phantom_use\` dummy** in plugin lib.rs. Reviewer pointed out dead-code "I'll need this later" stubs rot fast; Phase C will re-add the import in ~5 chars when it actually uses \`Manager\` for \`app.try_state::<...>()\`.
2. **Dropped \`Error::UnsupportedPlatform\` variant** since nothing constructs it in Phase B. Kept the empty \`Error\` enum + \`Result<T>\` alias so Phase C can grow the variant set without touching every command signature.
3. **Dropped the brittle \`$schema\` path** in \`permissions/default.json\` (\`../../../src-tauri/gen/schemas/...\`). The \`gen/\` directory is per-iOS-init and won't exist on a fresh CI checkout. The Tauri runtime doesn't read \`$schema\` — it's editor-only, and the autogenerated \`permissions/schemas/schema.json\` is the canonical reference.

### Documented (no change to behaviour)
4. **Effect coarseness** — added a comment block in \`session_timer.rs\` noting that the lifecycle Effect re-fires on every \`view_model\` push (any unrelated VM mutation triggers it), but the anchor-equality guard makes that idempotent. If we ever care, a \`Memo<Option<String>>\` over \`current_item_started_at\` would isolate the dependency.

- **Effect reactivity correctness** — Leptos tracks reads at \`.with()\` / \`.get()\` site equally. The closure return value isn't tracked; the source signal is. So the Effect re-runs on every \`view_model.set(...)\` from the Crux bridge.
- **Double-fire safety** — \`prev_anchor.set(...)\` is at the *end* of each match arm, and \`get_untracked()\` reads avoid creating a loop. Anchor-equality guard means re-runs with the same anchor fall through the catch-all arm.
- **\`links = "tauri-plugin-background-audio"\`** — required format for the \`tauri-plugin\` build script, must match \`package.name\`. Correct.
- **Tier-2-with-override** decision flagged in PR body.
- **Web bundle bloat** — JS shim is ~600 bytes uncompressed; trivial.
- **No e2e coverage added** — call sites are no-ops on web by design; nothing observable to test until Phase C and a physical device.

### Deferred (none from this PR)
No deferred items. The whole point of this PR is "scaffold the plumbing"; the actual deferred work is Phases C and D, which are tracked in the parent issue and the spec.

PR #450, opened 2026-05-07
feat(mobile): Swift impl for background-audio plugin (Phase C of #309)

jonyardley:
## Summary
**Phase C** of [intrada#309](https://github.com/jonyardley/intrada/issues/309) per [\`specs/background-audio-plugin.md\`](https://github.com/jonyardley/intrada/blob/main/specs/background-audio-plugin.md). Phase A ([intrada#431](https://github.com/jonyardley/intrada/pull/431), wall-clock timer) and Phase B ([intrada#441](https://github.com/jonyardley/intrada/pull/441), plugin scaffold) merged. This PR adds the actual Swift implementation: \`AVAudioSession\` activation, silent audio loop to keep the OS from suspending the app, \`MPNowPlayingInfoCenter\` for the lock screen, \`AVAudioSession.interruptionNotification\` for phone-call resume.

After this lands (and assuming on-device verification passes), the practice timer should keep running on the lock screen — the actual product blocker for shipping iOS to TestFlight.

**Marked draft** because lock-screen continuation, phone-call interruption recovery, and \`mixWithOthers\` behaviour can only be verified on a physical iPhone — the simulator's audio session model differs in ways that hide bugs (per CLAUDE.md "test on device" rule).

## What's in
- **\`BackgroundAudioPlugin.swift\`** (~300 lines) — \`Plugin\` subclass with three \`@objc\` methods. \`AVAudioSession.setCategory(.playback, .mixWithOthers)\`, silent loop via \`AVAudioPlayer\` (\`numberOfLoops=-1\`, volume=0 belt-and-braces), \`MPNowPlayingInfoCenter\` populated with title + artist (\`Item N of M\`) + elapsed + (best-effort) app-icon artwork. Synthetic 24h \`MPMediaItemPropertyPlaybackDuration\` so the lock-screen scrubber advances under \`playbackRate=1.0\` without us pushing updates every second. \`interruptionNotification\` handler recovers the session on \`.ended + .shouldResume\` (seeds Now Playing first, then reactivates so any reactivation lag is invisible).
- **All shared-state mutation hops to the main queue.** Interruption callbacks come in on a notification thread, and \`MPNowPlayingInfoCenter\` must be touched on main per Apple's docs. Single discipline (\`DispatchQueue.main.async\` at every entry point) avoids races.
- **\`silence.wav\`** — 1s 44.1kHz mono 16-bit silence, ~86KB, generated via Python's \`wave\`. Looped indefinitely to keep iOS from suspending the audio session.
[...]

PR #460, opened 2026-05-08
feat(mobile): Phase D — cleanup audit + Sentry for background-audio (#309)

jonyardley:
## Summary

- Move background-audio lifecycle Effect from \`<SessionTimer>\` up to \`<AuthenticatedApp>\` so \`end_session()\` fires on every session-end transition, regardless of which route is mounted
- Add Sentry breadcrumbs + error capture to the Rust plugin commands so we get real-device telemetry on bridge failures
- Defer in-app vm.error UI surfacing until we have failure data to cluster against (#458)

## Why this matters

Phase C (#450, just merged) shipped the Swift implementation, but the lifecycle Effect was still mounted inside \`<SessionTimer>\`. That meant the iOS AVAudioSession + silent loop would leak in two real-world flows:

1. User starts a session, taps a tab bar item or backs to home (without finishing) → \`<SessionTimer>\` unmounts → Effect disposed → audio session keeps running until the OS reclaims it.
2. User starts a session, navigates to \`/sessions/new\`, taps "Discard Session" → \`<SessionTimer>\` is not mounted at this route → Effect doesn't observe the \`Some → None\` transition → \`end_session()\` never fires.

Both leaks would have shipped to TestFlight without this fix. Mounting the Effect at \`<AuthenticatedApp>\` level fixes both because that owner outlives any route change while the user is signed in.

## What's intentionally not done

- **vm.error UI surfacing** — deferred to #458. We don't yet know which failure modes are user-actionable; telemetry-first.
- **Sign-out mid-session cleanup** — deferred to #459. Edge case; OS reclaims the session on app foreground.

## Test plan

- [x] \`cargo fmt --check\` passes
- [x] \`cargo clippy --workspace --exclude tauri-plugin-background-audio -- -D warnings\` passes
- [x] \`cargo clippy -p tauri-plugin-background-audio --target aarch64-apple-ios -- -D warnings\` passes (covers the iOS-gated breadcrumb/capture code that Linux CI doesn't see)
- [x] \`cargo test --workspace --exclude tauri-plugin-background-audio\` passes (418 tests)
- [x] \`cargo build -p intrada-mobile --target aarch64-apple-ios\` succeeds (host links plugin cleanly with new sentry dep)
- [ ] Device: 30-min session, lock + unlock, verify Now Playing still updates and timer is correct after unlock — folds into the Phase C device-test pass; not regressing what Phase C did.
- [ ] Device: start session → tap tab → wait → verify Now Playing card disappears and Control Center has no stale Intrada entry.
- [ ] Device: start session → navigate to /sessions/new → tap Discard → verify Now Playing card disappears within ~1s.

## Self-review

- ✅ Effect ownership: \`AuthenticatedApp\` lives as long as the user is signed in, strictly longer than any session.
- ✅ Crash recovery interaction: \`RecoverSession\` runs at line 148-153 of app.rs, Effect mounts on line ~163. Effect's first observation sees \`(None, Some(recovered))\` → fires \`begin_session()\` to seed lock-screen on relaunch with an in-progress session. Correct behaviour.
- ✅ Sentry version match: plugin (0.48) matches host (0.48), so \`Hub::current()\` lookup works across the FFI boundary.
- ✅ Breadcrumb on every command, capture only on Err — exactly what telemetry needs ("see context when something fails"), no noise on success path.
- ⚠️ Sign-out mid-session leaks the audio session — filed as #459, not blocker.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6823,7 +6823,9 @@ dependencies = [
 name = "tauri-plugin-background-audio"
 version = "0.0.1"
 dependencies = [
+ "sentry",
  "serde",
+ "serde_json",
  "tauri",
  "tauri-plugin",
  "thiserror 2.0.18",
--- a/crates/intrada-mobile/plugins/background-audio/Cargo.toml
+++ b/crates/intrada-mobile/plugins/background-audio/Cargo.toml
@@ -3,7 +3,7 @@ name = "tauri-plugin-background-audio"
 version = "0.0.1"
 edition = "2021"
 rust-version.workspace = true
-description = "Intrada-internal Tauri plugin: keeps the practice timer running while the app is backgrounded / the device is locked. Phase B scaffold — commands return Ok with no native side-effects yet."
+description = "Intrada-internal Tauri plugin: keeps the practice timer running while the app is backgrounded / the device is locked. Bridges to Swift on iOS for AVAudioSession + lock-screen Now Playing."
 # Required by tauri-plugin's build script (must match package.name).
 links = "tauri-plugin-background-audio"
 
@@ -11,6 +11,15 @@ links = "tauri-plugin-background-audio"
 tauri = { version = "2" }
 serde = { workspace = true }
 thiserror = { workspace = true }
+
+# iOS-only: sentry + serde_json. Used inside `#[cfg(target_os = "ios")]`
+# helpers to drop breadcrumbs / capture bridge errors against the host's
+# already-initialised Sentry hub (intrada-mobile/src-tauri/src/lib.rs).
+# Default sentry features off — the host owns transport (reqwest/rustls);
+# we only need the core API surface.
+[target.'cfg(target_os = "ios")'.dependencies]
+serde_json = { workspace = true }
+sentry = { version = "0.48", default-features = false }
 
 [build-dependencies]
 tauri-plugin = { version = "2", features = ["build"] }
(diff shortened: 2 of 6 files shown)


jonyardley:
**Self-review summary** (posting on the PR per CLAUDE.md so it's visible alongside CI, not buried in a subagent):

### What this changes
- `crates/intrada-web/src/app.rs` — call `mount_background_audio_lifecycle(view_model)` after init/recover, before view
- `crates/intrada-web/src/background_audio.rs` — extract Effect from `session_timer.rs` into a public function
- `crates/intrada-web/src/components/session_timer.rs` — strip the Effect block; replace with a comment pointing at the new home
- `crates/intrada-mobile/plugins/background-audio/src/lib.rs` — add iOS-gated `breadcrumb()` + `capture_bridge_error()` helpers, call from each command
- `crates/intrada-mobile/plugins/background-audio/Cargo.toml` — add `sentry` + `serde_json` as iOS-only deps

### Things I checked
- ✅ The Effect's first run on mount observes `view_model` after `RecoverSession` has already populated it, so crash-recovery on a locked phone seeds the lock screen correctly.
- ✅ Cargo `target.'cfg(target_os = "ios")'.dependencies` is the right shape — verified by building both the iOS target (`aarch64-apple-ios`) and the host target without warnings.
- ✅ `breadcrumb` data uses `serde_json::Map` → `BTreeMap<String, Value>` collect; sentry 0.48 accepts this directly.
- ✅ No `cargo fmt`/clippy/test failures locally (workspace + iOS-target plugin both clean).

### Things I flagged but didn't fix
- vm.error UI surfacing — #458 (telemetry-first, decide what's user-actionable later)
- Sign-out during active session — #459 (edge case, OS reclaims)

### What I can't verify locally
The actual cleanup behaviour (the leaks this PR fixes) is only observable on a physical iOS device with the lock-screen Now Playing card visible. Folds into the Phase C device-test pass.

Issue #477, opened 2026-05-08
MCP server — agent-callable API


jonyardley:
Spec: [`specs/mcp-server.md`](../blob/main/specs/mcp-server.md) — Tier 3 per CLAUDE.md.

## Why

intrada has rich practice data — items, sets, sessions with scoring, weekly analytics — and no agent-friendly way to interact with it. An MCP server exposes the API as tools an LLM client (Claude Desktop, Cursor, custom agents, web-based agents) can call on the user's behalf — read their library, populate it from natural-language prompts, summarise practice patterns, and feed structured context for advice.

Cheapest concrete step into Layer 5 (Guide) territory: the AI advice surface lives in the user's existing chat client, not in a new in-app feature.

## Phases (each its own PR / sub-issue)

- [ ] **Phase 1 — Service-layer refactor.** Extract per-route business logic so HTTP routes and MCP tools share the same code. Pure refactor; existing tests prove it.
- [ ] **Phase 2 — PAT issuance.** `mcp_tokens` table + `/account/tokens` CRUD + account-settings UI (Pencil first, web + iOS).
- [ ] **Phase 3 — MCP server, reads only.** New `intrada-mcp` crate at `/mcp`. List/get tools + `get_practice_summary` wide view. Agent can already give meaningful advice on read data alone — useful checkpoint.
- [ ] **Phase 4 — Write tools.** `create_item`, `update_item`, `delete_item`, `create_set`, `update_set`, `bulk_import_items` with `dry_run` confirmation. `mcp_audit_log` table + UI.
- [ ] **Phase 5 — OAuth 2.1 + DCR.** OAuth server in intrada-api, authorize page on intrada-web, DCR endpoint. Existing PATs keep working.
- [ ] **Phase 6 — Polish.** Rate limits (60 req/min/token, 429), telemetry (Sentry spans), docs.

## Key decisions (from spec)

- PAT in v1, OAuth in v2 — schema supports both.
- Lives inside intrada-api, not a separate service.
- No `get_advice` tool — agents do the advising on top of structured data.
- Bulk writes use `dry_run` previews; single writes are reversible + audited.
- Wire protocol uses API terminology (`items`, `sets`); tool descriptions use user-facing language for the agent.
- Single flat scope for v1; `mcp_tokens` schema leaves room for granular scopes later.
- Settings-only discoverability (no dashboard CTA).

## Open questions

- iOS PAT-entry UX (cross-device copy/paste) — Pencil first during Phase 2.
[...]

PR #522, opened 2026-05-08
feat(api): MCP server (reads-only) at /api/mcp (Phase 3 of #477)

jonyardley:
## Summary

Phase 3 of the [MCP server epic](https://github.com/jonyardley/intrada/issues/477). The actual MCP server. After this PR, an MCP client (Claude Desktop, Cursor, custom agents) can:

1. Connect to \`POST /api/mcp\` with a PAT bearer
2. Call \`initialize\` → get server info + capabilities
3. Call \`tools/list\` → discover the 7 reads-only tools
4. Call \`tools/call\` → invoke any tool

Resolves [#481](https://github.com/jonyardley/intrada/issues/481) (CORS policy decision lands here).

## Tools

| Name | Args | Returns |
|------|------|---------|
| \`list_items\` | \`kind?: \"piece\" \| \"exercise\"\` | All items, optionally filtered |
| \`get_item\` | \`id\` | Item with composer/key/tempo/notes/tags |
| \`list_sets\` | (none) | Routines |
| \`get_set\` | \`id\` | Routine with ordered entries |
| \`list_sessions\` | \`start?\`, \`end?\` (RFC 3339) | Sessions in window |
| \`get_session\` | \`id\` | Session with per-item scores/notes |
| \`get_practice_summary\` | \`start\`, \`end\` | Pre-joined wide view: total minutes, items practiced, score trends, per-item breakdown |

\`get_practice_summary\` is the spec's killer wide-view tool — saves the agent 10 round trips for typical \"how was my week\" questions.

## Design decisions

- **Module inside intrada-api, not a separate crate.** A standalone crate would create a circular dep (it needs \`AppState\`/\`AuthUser\`/\`services::*\`). The spec's rationale for \"separate crate\" was deployment-related, which we get either way.
- **Hand-rolled JSON-RPC, not rmcp.** Per-user state through rmcp's session-factory closure is awkward and we don't need SSE/streaming for reads. Wire protocol is the same; swap is a focused refactor if we want streaming later.
- **Tool errors → \`isError: true\` in JSON-RPC \`result\`** (per MCP spec). JSON-RPC errors are reserved for protocol-level problems (parse, method not found, invalid params).
- **Permissive CORS for /api/mcp/* only**, strict CORS preserved for the existing /api/* surface. Per [#481](https://github.com/jonyardley/intrada/issues/481).
- **Tool descriptions use user-facing language** (\"piece\", \"exercise\", \"routine\") even though wire-protocol field names match the API (\`items\`, \`sets\`).

## Test plan

- [x] 10 new integration tests in \`tests/mcp_test.rs\` (initialize, tools/list, tools/call happy + filter + unknown id + unknown tool, unknown method, invalid jsonrpc, notification → 204, get_practice_summary aggregation)
- [x] \`cargo test -p intrada-api\` — 101 passed (was 91)
- [x] \`cargo fmt --check\` clean
- [x] \`cargo clippy --workspace --exclude intrada-mobile --exclude tauri-plugin-background-audio -- -D warnings\` clean
- [ ] CI on this PR

## Manual smoke test

\`\`\`bash
TOKEN=$(curl -sX POST localhost:3001/api/account/tokens \\
  -H 'content-type: application/json' \\
  -d '{\"name\":\"local-mcp\"}' | jq -r .token)

# initialize
curl -sX POST localhost:3001/api/mcp \\
  -H \"Authorization: Bearer $TOKEN\" \\
  -H 'content-type: application/json' \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"initialize\",\"id\":1}' | jq

# tools/list
curl -sX POST localhost:3001/api/mcp \\
  -H \"Authorization: Bearer $TOKEN\" \\
  -H 'content-type: application/json' \\
  -d '{\"jsonrpc\":\"2.0\",\"method\":\"tools/list\",\"id\":2}' | jq
\`\`\`

## Out of scope (Phase 4 etc.)

- Write tools (create_item, bulk_import_items with dry_run, etc.) — Phase 4.
- Audit log (\`mcp_audit_log\`) — Phase 4. The \`AuthSource::Pat { token_id }\` from #501 is already plumbed through; Phase 4's audit handlers can read it.
- OAuth 2.1 + DCR — Phase 5.
- Rate limits — Phase 6.

## Closing

- Closes [#481](https://github.com/jonyardley/intrada/issues/481) (CORS decision implemented).

### DIFF
--- a/crates/intrada-api/src/lib.rs
+++ b/crates/intrada-api/src/lib.rs
@@ -2,6 +2,7 @@ pub mod auth;
 pub mod clerk;
 pub mod db;
 pub mod error;
+pub mod mcp;
 pub mod migrations;
 pub mod routes;
 pub mod services;
--- a/crates/intrada-api/src/mcp/handlers.rs
+++ b/crates/intrada-api/src/mcp/handlers.rs
@@ -166,7 +166,11 @@ pub async fn get_practice_summary(
         .collect();
 
     let sessions_count = in_window.len();
-    let total_minutes: u64 = in_window.iter().map(|s| s.total_duration_secs / 60).sum();
+    // Sum seconds first, divide once at the end — dividing per-session
+    // would drop sub-minute remainders. With 10 × 90s sessions the
+    // per-session approach reads as 10min instead of the correct 15.
+    let total_secs: u64 = in_window.iter().map(|s| s.total_duration_secs).sum();
+    let total_minutes = total_secs / 60;
 
     let mut item_ids: std::collections::HashSet<String> = std::collections::HashSet::new();
     let mut score_sum: u32 = 0;
--- a/crates/intrada-api/src/mcp/handlers.rs
+++ b/crates/intrada-api/src/mcp/handlers.rs
@@ -1,0 +1,253 @@
+//! Tool implementations.
+//!
+//! Each handler:
+//! - Takes the typed arguments parsed from the JSON-RPC `arguments` object.
+//! - Calls into `crate::services::*` (the same layer HTTP routes use).
+//! - Returns the result as `serde_json::Value` for the dispatcher to wrap
+//!   in MCP's `content: [{ type: "text", text: "<JSON>" }]` shape.
+//!
+//! Date-range filtering on `list_sessions` happens here rather than in the
+//! service layer to keep this PR scoped — pushing the filter down is a
+//! follow-up refactor.
+
+use chrono::{DateTime, Utc};
+use libsql::Connection;
+use serde::Deserialize;
+use serde_json::{json, Value};
+
+use intrada_core::domain::item::ItemKind;
+
+use crate::error::ApiError;
+use crate::services;
+
+// ── list_items ─────────────────────────────────────────────────────────
+
+#[derive(Debug, Deserialize, Default)]
+pub struct ListItemsArgs {
+    #[serde(default)]
+    pub kind: Option<String>,
+}
+
+pub async fn list_items(
+    conn: &Connection,
+    user_id: &str,
+    args: ListItemsArgs,
+) -> Result<Value, ApiError> {
+    let kind_filter = match args.kind.as_deref() {
+        None => None,
+        Some("piece") => Some(ItemKind::Piece),
+        Some("exercise") => Some(ItemKind::Exercise),
+        Some(other) => {
+            return Err(ApiError::Validation(format!(
+                "Invalid kind: {other:?} (expected \"piece\" or \"exercise\")"
+            )))
+        }
+    };
+    let items = services::items::list_items(conn, user_id).await?;
+    let filtered: Vec<_> = items
+        .into_iter()
+        .filter(|i| kind_filter.is_none() || Some(i.kind.clone()) == kind_filter)
+        .collect();
+    serde_json::to_value(filtered).map_err(|e| ApiError::Internal(format!("serialize items: {e}")))
+}
+
+// ── get_item ───────────────────────────────────────────────────────────
+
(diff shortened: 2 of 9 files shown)


jonyardley:
**Approve with comments.** Solid Phase 3 — clean module split, hand-rolled JSON-RPC reads naturally, test coverage good for protocol behaviour. Verified locally: `cargo test -p intrada-api --test mcp_test` → 10 passed; `cargo fmt --check` clean; `cargo clippy --workspace --exclude intrada-mobile --exclude tauri-plugin-background-audio -- -D warnings` clean. CORS split, auth integration, revoked-PAT rejection path, and JSON-RPC error semantics all check out.

### Important
- **`get_practice_summary` total_minutes precision bug** (`mcp/handlers.rs:169` and `:225`). Per-session/per-item integer division before summing drops sub-minute remainders every time. 10 sessions of 90s read as 10 min (real: 15); 10 sessions of 50s read as 0 min (real: ~8). The existing test uses 1800/600/1200s — all clean multiples of 60 — so it doesn't exercise the lossy path. Fix: sum seconds first, divide once at the end (`secs.sum() / 60`). Add a regression test with a 90s entry. This is the "killer wide-view tool"; numbers being silently low erodes agent trust fast.

### Nits
- **`AuthSource::Pat { token_id }` discarded with `..`** in `server.rs:34`. Add `// NOTE(#477 phase 4): token_id read here for mcp_audit_log` so the next reader knows where audit logging plugs in. Cheap signposting.
- **`routes/mod.rs:65` comment "Order matters: longest-prefix wins"** — slightly misleading. Path specificity wins regardless of registration order in axum routers. Either drop "Order matters" or rephrase as "Path specificity routes `/api/mcp/*` here…".
- **`list_sessions` date filter on `started_at` only** — a session that starts inside the window but completes after gets included; one starting just before the window but covering it gets excluded. Probably fine for the agent UX, but the tool description (`tools.rs:75`) doesn't mention which timestamp drives the filter. Two-line clarification would help the LLM phrase user-facing summaries correctly.
- **Tool descriptions** otherwise read well — "piece"/"exercise"/"routine" terminology consistent, schemas tight, `additionalProperties: false` everywhere. `get_practice_summary` description is the strongest sell.
- `parse_and_run`'s null-args normalisation (`server.rs:238-242`) is a nice ergonomic touch. Worth a one-line comment that this is why agents don't have to send `"arguments": {}` for `list_sets`.

### Test gaps (consider, not blocking)
- No direct test that PAT auth works against `/api/mcp` (transitively covered by `tokens_test.rs::pat_authenticates_subsequent_requests`, but a one-liner asserting `Bearer intrada_pat_…` reaches MCP would be cheap insurance).
- No CORS preflight OPTIONS test against `/api/mcp` confirming permissive headers (vs strict). Given that's the headline decision from #481, an explicit assertion seems worth ~10 lines.
- No revoked-PAT-against-MCP test. Auth extractor tests cover it generically; one MCP-specific case would close the loop.

All three could ship as a follow-up issue rather than blocking this PR.

### Out of scope (open as separate issues)
- Sub-minute precision regression test + push date-range filtering into `services::sessions::list_sessions` (mentioned in `handlers.rs:9-11` already). Single follow-up issue.
- Direct PAT/CORS/revoked-PAT MCP test trio. Single follow-up issue.

Module-not-crate decision is correct given AppState/AuthUser/services coupling — separate crate would force extracting all of those for marginal benefit. Hand-rolled vs rmcp also sound for reads-only with per-user context; the `tower::Layer` workaround is possible but adds complexity without buying us streaming/SSE we don't need. Both decisions read as deliberate, not lazy.

PR #537, opened 2026-05-08
fix(web): persist focus-mode preference per session + diagonal-arrow icons

jonyardley:
## Summary

Two related practice-flow polish items reported together.

### 1. Focus mode no longer auto-snaps back on
Today every mount of `/sessions/active` forces `focus_mode=true`, every unmount forces `false`. So: start a session (focus on, by design), exit focus, navigate away, come back — and you're slammed into focus mode again, overriding the user's choice.

Now we remember the toggle keyed by \`ActiveSessionView::started_at\` (RFC3339, unique per session). Re-entering the same session restores the user's last toggle. Fresh sessions still default to focus-on. When the session ends (no \`active_session\`), the pref clears so the next session starts fresh.

Implementation: new \`SessionFocusPref(RwSignal<Option<(String, bool)>>)\` context provided at app level; \`SessionActiveView\` reads it on mount and writes it on \`on_cleanup\`.

### 2. Diagonal-arrow icons for the focus toggle
\`ChevronUp/Down\` read as \"expand/collapse list\" — doesn't match what the button actually does (toggle full-screen focus mode). Swapped for Lucide-style \`maximize-2\` / \`minimize-2\` (the diagonal-arrow pair) which is the conventional fullscreen toggle iconography.

## Test plan
- [x] Start a session → lands in focus mode with **Minimize** icon top-right (header + tab bar hidden).
- [x] Click Minimize → focus off, **Maximize** icon, header + tab bar return.
- [x] Navigate to Library → tap Practice tab → returns to \`/sessions/active\` with focus still **off** (this is the bug fix).
- [x] Toggle back on, navigate away, return → focus still **on**.
- [ ] Verify on iOS simulator that haptics still fire correctly on toggle.

Verified in-browser end-to-end via dev preview. \`cargo check\` / \`cargo clippy -p intrada-web -- -D warnings\` / \`cargo fmt --check\` all clean locally.

### DIFF
--- a/crates/intrada-web/src/app.rs
+++ b/crates/intrada-web/src/app.rs
@@ -53,6 +53,15 @@ pub struct AuthState {
     pub auth_error: RwSignal<bool>,
 }
 
+/// Remembers the user's focus-mode toggle for the currently-active session,
+/// keyed by `ActiveSessionView::started_at` (RFC3339, unique per session).
+///
+/// Re-entering the active-session route looks this up so a user who
+/// explicitly exited focus mode mid-session isn't snapped back into it on
+/// return. Cleared when the session ends so the next session starts fresh.
+#[derive(Clone, Copy)]
+pub struct SessionFocusPref(pub RwSignal<Option<(String, bool)>>);
+
 #[component]
 pub fn App() -> impl IntoView {
     // Auth state signals — drive the auth gate
@@ -143,6 +152,7 @@ pub fn App() -> impl IntoView {
     let is_loading = IsLoading::new(true);
     let is_submitting = IsSubmitting::new(false);
     let focus_mode = FocusMode(RwSignal::new(false));
+    let session_focus_pref = SessionFocusPref(RwSignal::new(None));
 
     provide_context(auth);
     provide_context(core.clone());
@@ -150,6 +160,7 @@ pub fn App() -> impl IntoView {
     provide_context(is_loading);
     provide_context(is_submitting);
     provide_context(focus_mode);
+    provide_context(session_focus_pref);
     provide_toast();
 
     // Session-lifecycle Effect (#309 Phase D + #474 Phase B). Drives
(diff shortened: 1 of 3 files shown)


jonyardley:
## Self-review (\`superpowers:code-reviewer\`)

**Verdict: ship it.** No blockers, no important findings. Applied one inline nit.

### Confirmed correct
- **Persistence logic.** \`view_model\` provided at App-level outlives every route, so \`with_untracked\` in \`on_cleanup\` is safe — no torn-down signal hazard. \`get_untracked\` everywhere prevents accidental subscriptions / Effect-loop risk.
- **Storage shape.** \`Option<(String, bool)>\` is right for "single keyed pref". HashMap overkill; two signals split state; localStorage unwarranted given session lifetime. Endorsed.
- **Lifecycle correctness.** Discard / Summary / End → \`session_status\` ≠ Active → component unmounts → cleanup with \`started_at\` derived from current view_model. If session is gone, pref is cleared. If session is alive, pref is saved keyed to its id. Next session (different \`started_at\`) defaults to focus-on. No stale-pref ghost.
- **Icon paths.** \`<polyline>\` / \`<line>\` are already used elsewhere (Clock, Star, Ban, Calendar), and matching Lucide source verbatim makes future verification easier. Keeping primitives as-is.
- **A11y.** \`aria-label\` flips Enter ↔ Exit on toggle (unchanged); icon stays \`aria-hidden\`; semantic carried by the label. No regression.

### Applied inline
- **Comment** clarifying that the trailing \`focus_mode.set(false)\` in cleanup is still needed because \`FocusMode\` is app-level — chrome (header/tab bar/footer) won't reappear off-route otherwise. Prevents a future "why is this still here?" investigation. Pushed as a follow-up commit.

### Deferrable (not applied)
- \`SessionFocusPref\` has no \`get\`/\`set\` helpers like \`FocusMode\` does. Two callsites only — adding helpers would be ceremony for ceremony's sake. Add if a third caller appears.

### Local checks
- \`cargo check -p intrada-web\` ✅
- \`cargo clippy -p intrada-web -- -D warnings\` ✅
- \`cargo fmt --check\` ✅
- End-to-end browser test: focus persists in both directions across Library → Practice round-trip on the same session ✅

PR #579, opened 2026-05-09
fix(web): show "signing you in" on /sso-callback instead of 404

jonyardley:
## Summary

A friend on Android Chrome reported a 404 when signing in with Google — saw "Page Not Found" on `myintrada.com/sso-callback` and gave up. Sentry confirms sign-in actually succeeded (`user_3DTn59RtFvTHZYK9yaP7v9SsjHb` reached `/library` 44s later) — the 404 was a UX flicker, not a real failure.

**Root cause**: there was no Leptos route for `/sso-callback`. The inline JS bridge in `index.html` handles the OAuth callback, but it has to download `@clerk/clerk-js` from jsdelivr.net first (5-15s on slow mobile), then call `clerk.handleRedirectCallback()`. Meanwhile the WASM bundle hydrates faster and Leptos renders `NotFoundView` because no route matches. Once Clerk completes, it redirects to `/` — but by then the user has already screenshotted the 404.

**Fix**: add a real `/sso-callback` route that renders "Signing you in…" with a spinner, plus a 15s timeout fallback to a "Back to sign in" link. The view also watches `AuthState.is_authenticated` and pushes to `/library` itself when the listener flips — defence-in-depth in case Clerk's own redirect fails.

Edge cases handled (per self-review):
- One-shot navigated guard prevents redirect loops if `is_authenticated` flickers
- `on_cleanup` cancels the 15s timeout so it can't write to a disposed signal
- Landing on `/sso-callback` with no query params (manual URL entry, refresh after success) shows the friendly fallback immediately instead of waiting 15s

The actual auth wiring in [index.html:170-275](crates/intrada-web/index.html#L170-L275) is unchanged.

## Test plan
- [ ] CI: `cargo fmt --check`, `cargo test`, `cargo clippy -- -D warnings` (locally clean against wasm32 target)
- [ ] Visual: navigate to `/sso-callback` directly while signed out — verify "Signing you in…" spinner instead of 404 (defence-in-depth Effect will redirect to `/library` if signed in)
- [ ] Visual: navigate to `/sso-callback` with no query params — verify friendly fallback ("This page is part of signing in. Take me back…") shows immediately
- [ ] Real device: Android Chrome on cellular, sign in with Google — verify spinner stays visible during Clerk SDK download, no 404 flicker

## Out of scope (deferred)
- Sentry noise from Leptos `transition=true` View Transitions API (`AbortError` / `InvalidStateError`, DOMException 20/11) — cosmetic, not load-bearing
- Bigger refactor: drop the runtime CDN load of `@clerk/clerk-js` in `index.html` — bundle it or `modulepreload` so the SDK is ready before the OAuth callback lands
- Factor the inline `animate-spin` SVG combo (used here, in `button.rs`, and `design_catalogue.rs`) into a `Spinner` primitive

### DIFF
--- a/crates/intrada-web/src/app.rs
+++ b/crates/intrada-web/src/app.rs
@@ -22,7 +22,7 @@ use crate::views::{
     AccountDeleteView, AddLibraryItemForm, AnalyticsPage, DetailView, EditLibraryItemForm,
     LibraryListView, LoginView, McpAuditView, McpTokensView, NotFoundView, OAuthConsentView,
     SessionActiveView, SessionNewView, SessionSummaryView, SessionsAllView, SessionsListView,
-    SetDetailView, SetEditView, SettingsView, WelcomeView,
+    SetDetailView, SetEditView, SettingsView, SsoCallbackView, WelcomeView,
 };
 use intrada_web::core_bridge::{init_core, load_session_in_progress, process_effects};
 use intrada_web::js_bridge;
@@ -226,6 +226,7 @@ fn AppRoutes() -> impl IntoView {
             // ─── Public routes ────────────────────────────────────────
             <Route path=path!("/") view=|| view! { <WelcomeView /> } />
             <Route path=path!("/login") view=|| view! { <LoginView /> } />
+            <Route path=path!("/sso-callback") view=|| view! { <SsoCallbackView /> } />
 
             // ─── Private routes ───────────────────────────────────────
             // /library/new MUST come before /library/:id to avoid "new"
--- a/crates/intrada-web/src/views/mod.rs
+++ b/crates/intrada-web/src/views/mod.rs
@@ -19,6 +19,7 @@ pub mod sessions_all;
 pub mod set_detail;
 pub mod set_edit;
 pub mod settings;
+pub mod sso_callback;
 pub mod welcome;
 
 pub use account_delete::AccountDeleteView;
@@ -42,4 +43,5 @@ pub use sessions_all::SessionsAllView;
 pub use set_detail::SetDetailView;
 pub use set_edit::SetEditView;
 pub use settings::SettingsView;
+pub use sso_callback::SsoCallbackView;
 pub use welcome::WelcomeView;
--- a/crates/intrada-web/src/views/sso_callback.rs
+++ b/crates/intrada-web/src/views/sso_callback.rs
@@ -1,0 +1,116 @@
+use leptos::prelude::*;
+use leptos_router::components::A;
+use leptos_router::hooks::use_navigate;
+use leptos_router::NavigateOptions;
+use wasm_bindgen::closure::Closure;
+use wasm_bindgen::JsCast;
+
+use crate::app::AuthState;
+
+/// Landing route for the Clerk OAuth redirect — `/sso-callback`.
+///
+/// The actual callback is processed by the inline JS bridge in `index.html`
+/// (it calls `clerk.handleRedirectCallback` once the SDK finishes loading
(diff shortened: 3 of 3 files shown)


jonyardley:

Ran before pushing per CLAUDE.md (auth-domain Tier 2 → required). Findings + how they were addressed:

### Blockers (fixed inline before push)
1. **Redirect loop guard** — `Effect` had no one-shot guard; `is_authenticated` flickering true→false→true would re-fire `navigate("/library")`. Added `RwSignal<bool>` `navigated` flag, set on first fire ([sso_callback.rs:38-49](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/sso_callback.rs#L38-L49)).
2. **Stale signal write after dispose** — `spawn_local`/`TimeoutFuture` outlives the view's owner; would log warnings or panic when writing `timed_out` after dispose. Switched to `set_timeout_with_callback_and_timeout_and_arguments_0` returning a handle stored in `RwSignal<Option<i32>>`, with `on_cleanup` calling `clear_timeout_with_handle` ([sso_callback.rs:53-77](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/sso_callback.rs#L53-L77)). Pattern matches `context_menu.rs:107-121`.

### Important (fixed inline)
3. **Manual URL entry edge case** — landing on `/sso-callback` with no `?code=` (refresh after success, deep link, manual URL) would sit on the spinner for 15s then show "We couldn't finish signing you in" — misleading copy. Now detects empty query string synchronously on mount and shows "This page is part of signing in. Take me back" immediately ([sso_callback.rs:28-32, 92-100](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/sso_callback.rs#L28-L32)).
4. **Route placement** — confirmed correct (public, alongside `/` and `/login`). `AuthenticatedShell` would redirect to `/login`, hijacking the callback.
5. **A11y on timeout fallback** — added `role="alert"` to the fallback `<p>` and `aria-live="polite"` on the outer container so screen readers announce the spinner→error swap ([sso_callback.rs:82, 96](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/sso_callback.rs#L82)).

### Nice-to-have (deferred, not blocking)
6. **Spinner primitive** — the inline `animate-spin` SVG combo here matches duplicates in [button.rs:121](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/components/button.rs#L121) and [design_catalogue.rs:1589](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/design_catalogue.rs#L1589). Factoring a shared `<Spinner />` primitive is a separate cleanup PR; out of scope here.
7. **Link-style class duplication** — `inline-flex items-center gap-2 text-accent-text hover:text-accent-hover font-medium` duplicates the shape used in [not_found.rs:14](https://github.com/jonyardley/intrada/blob/40e56f2/crates/intrada-web/src/views/not_found.rs#L14). Kept identical for visual parity until/unless we factor a shared primitive.

### Verification
- `cargo fmt --check` ✓
- `cargo check -p intrada-web --target wasm32-unknown-unknown` ✓
- `cargo clippy -p intrada-web --target wasm32-unknown-unknown -- -D warnings` ✓
- WASM bundle inspection confirms all three copy strings are present ("Signing you in…", "Back to sign in", "couldn't finish signing you in")
- Defence-in-depth Effect verified end-to-end (auto-redirected to `/library` when `is_authenticated=true` on mount)

PR #634, opened 2026-05-10
fix(mobile,web,api): unbreak iOS practice session — Bundle.module crash + Tauri args + libsql tls

jonyardley:
## Summary

Three fixes that, taken together, restore a working iOS practice session:

1. **`fix(mobile)` — Bundle.module crash** *(the headline bug)*
   `BackgroundAudioPlugin` crashed the app with `EXC_BREAKPOINT` on first session start because Tauri 2's iOS build pipeline doesn't link the SwiftPM resource bundle that `Bundle.module` synthesizes against. Drop the resource entirely; synthesize 1 second of silent PCM inline (~16KB static `Data`) and feed `AVAudioPlayer(data:)`. No more `Bundle.module`, no more dependence on Tauri's iOS resource handling.

2. **`fix(web)` — Tauri plugin args wrapping**
   Both plugin JS bridges sent payload fields flat (`{ item_title: "..." }`), but Tauri 2's `#[tauri::command]` macro deserializes JSON keyed by the *parameter name*. Since both plugins declare `args: BeginArgs` (etc.), the correct shape is `{ args: { item_title: "..." } }`. Without this every command call was rejected with `missing required key args` — the silent-loop activator and Live Activity were silent no-ops.

3. **`fix(api)` — restore libsql `tls` feature**
   PR #623 set `default-features = false` on libsql to slim the build but dropped the `tls` feature in the process, breaking remote Turso connectivity. Adds `"tls"` back to the feature list.

The three are independent root causes that surfaced in sequence while debugging the iOS Live Activity / background-audio path. Kept as three commits for clean review and easy individual revert.

## Crash incident

`8470A7CD-46C8-44BD-83F3-51784F12A690` — `EXC_BREAKPOINT` on `com.apple.main-thread` inside `_assertionFailure ← NSBundle.module init ← BackgroundAudioPlugin.startSilentLoop`. iOS 26.4.2.

## Verification (on-device, iPhone)

- App no longer crashes on session start
- Lock Screen Live Activity card renders + ticking timer
- Dynamic Island compact representation appears
- Item advance updates the widget
- Session end dismisses the widget within ~1s (`dismissalPolicy: .immediate`)
- Background timer continuity works (silent loop is doing its job — screen locked, returned 2-3 min later, timer correct)
- `.mixWithOthers` allows other audio (Sonos backing track) to keep playing without ducking

## Test plan

- [x] `cargo fmt --check` clean
- [x] `cargo clippy --workspace --exclude tauri-plugin-background-audio --exclude tauri-plugin-live-activity -- -D warnings` clean
- [x] `cargo clippy -p tauri-plugin-background-audio --target aarch64-apple-ios -- -D warnings` clean
- [x] On-device verification (4 paths above)
- [ ] CI green
- [ ] Verify in Web Inspector that `unhandled rejection` no longer appears for plugin invokes when running iOS dev build

## Notes for reviewer

- The inline WAV synthesis in `BackgroundAudioPlugin.swift` is a drop-in for an 86KB `silence.wav` file. Output is the canonical 44-byte WAV header + 16,000 zero bytes (8kHz mono 16-bit PCM, 1s). Verified on device that `AVAudioPlayer(data:)` accepts it and `prepareToPlay()` returns true.
- `Package.swift` no longer declares `resources:`. Without that declaration SwiftPM doesn't emit `Bundle.module` at all, so any future accidental reference to it is now a compile error rather than a runtime fatalError. This is intentional belt-and-braces.
- The libsql change is a regression fix for #623 and could ship independently; bundling for a single review pass given the same overall debug session.


jonyardley:

**Bottom line: ship it.** Three correct, focused fixes. WAV header math verified byte-for-byte. `args` wrapping verified against the Rust command macro source. iOS clippy clean.

### Blockers
None.

### Important findings (deferred — see follow-up issues)

1. **Commit body for `11acef8` is slightly mis-located.** It says "Both background-audio and live-activity plugins declare their command parameters as a single `args: BeginArgs`" — this could be misread as the *Swift* `parseArgs(...)` requiring the wrapping. It's actually the **Rust** command macro at e.g. [`crates/intrada-mobile/plugins/background-audio/src/lib.rs:101`](crates/intrada-mobile/plugins/background-audio/src/lib.rs#L101) (`async fn begin_session(app, args: BeginSessionArgs)`) that requires the JS to send `{ args: { ... } }`. The Swift `parseArgs` then sees the inner struct flat. If anyone later renames the **Rust** parameter (`args` → `payload`), the JS bridges need to match the Rust parameter name. Not amending the commit (would require a force-push); flagging here so the PR thread carries the corrected mental model.

2. **The libsql `tls` restore re-introduces a TLS dual-stack** that PR #623 just consolidated away. `Cargo.lock` regrows by 11 packages (rustls, rustls-native-certs, tokio-rustls, webpki-roots, hyper-rustls, etc.) on top of the existing native-tls path used by reqwest / rust-s3 / sentry / jsonwebtoken. The fix is correct (Turso requires TLS), but worth a follow-up to audit whether libsql 0.9 exposes a native-tls transport variant. → [intrada#639](https://github.com/jonyardley/intrada/issues/639)

3. **No automated test would catch a regression of any of these three bugs.** Two of them are cheaply addressable: a serde round-trip test for the plugin args wire shape, and a Swift unit test on the synthesized `silenceWav` byte count + `RIFF`/`WAVE` magic. The libsql TLS regression is harder to test cheaply — out of scope. → [intrada#641](https://github.com/jonyardley/intrada/issues/641)

### Suggestions

4. After this PR, the only realistic failure mode for `startSilentLoop()`'s `AVAudioPlayer(data:)` call is a real bug (the bytes are static and deterministic). Mirror the Rust-side `capture_bridge_error` pattern and forward the error to Sentry in addition to `Logger.error`, so we'd actually see it in production if it ever fired. → [intrada#640](https://github.com/jonyardley/intrada/issues/640)

5. **Tier sensitivity (CLAUDE.md)** — this PR touches Tauri plugin IPC contracts and DB connectivity, which CLAUDE.md flags for tier promotion. In practice the work was minimum-viable correctness restoration on already-implemented systems, so the bug-fix tier was right. Noting for future bug-vs-feature judgement.

### What was verified

- `cargo fmt --check` — clean
- `cargo clippy --workspace --exclude tauri-plugin-background-audio --exclude tauri-plugin-live-activity -- -D warnings` — clean
- `cargo clippy -p tauri-plugin-background-audio --target aarch64-apple-ios -- -D warnings` — clean
- WAV header math: `blockAlign = 2`, `byteRate = 16000`, `dataSize = 16000`, `riffSize = 16036` (file size − 8 per RIFF spec) ✓, total file = 44 + 16000 = 16044 bytes ✓
- `args` wrapping covers every command call that takes parameters; `end` / `end_session` correctly omit it (no parameters)
- No other code path in `crates/intrada-mobile/` references `Bundle.module` for the background-audio plugin
- On-device verification (iPhone iOS 26.4.2): no crash, Live Activity renders + ticks + dismisses, item advance updates, background timer continuity through screen lock, `.mixWithOthers` allows other audio (Sonos) to keep playing

PR #654, opened 2026-05-10
fix(ux): splash flicker, reflection card, focus width

jonyardley:
## Summary

- **Splash logo flicker** — added `letter-spacing: -0.02em` to the HTML splash inline style to match the `page-title` CSS class, eliminating the subtle text shift during fade-out
- **Reflection sheet shows current item** — added `current_item_title` / `current_item_type` props so the sheet displays the item being reflected on as the primary heading, with "Up next:" as a compact secondary line
- **Focus mode width** — removed `align-items: center` from `focus-mode-container` so content stretches to full container width, matching non-focus mode

## Test plan

- [ ] Launch app on iOS — verify splash fades without any text shift
- [ ] Start a multi-item session, tap Next Item — verify the reflection sheet heading shows the item just completed (not the next one)
- [ ] Verify "Up next:" line appears below the form with the next item name
- [ ] On the last item, verify no "Up next:" line appears
- [ ] Toggle focus mode during a session — verify content width stays consistent


jonyardley:
## Self-review

**Scope**: Tier 1 UX fix — 5 files, +46/−28 lines. All three fixes are narrowly scoped.

### Fix 1: Splash letter-spacing ✅
Inline style now includes `letter-spacing: -0.02em`, matching the `page-title` utility class exactly. Minimal, correct.

### Fix 2: Reflection card shows current item ✅
Two new required props (`current_item_title`, `current_item_type`) on `ItemReflectionSheet`. The old "Up Next" heading is replaced by the current item name + type badge. "Up next:" renders as a compact secondary line below the form, only when there is a next item. Design catalogue demo updated with the new props.

### Fix 3: Focus mode width ✅
Removed `align-items: center` from `focus-mode-container`. Children now default to `stretch`, matching non-focus mode width. Container keeps `justify-content: center` for vertical centering.

**No issues found.** `cargo check`, `cargo fmt --check`, and `cargo clippy` all pass.

PR #805, opened 2026-06-01
Native iOS: source Library filter + totals from the ViewModel (#792)

jonyardley:
Closes #792.

## Problem
The Library filter pill (All/Pieces/Exercises) was held as local SwiftUI `@State` in `LibraryScreen` and re-asserted onto the core via `.onAppear` — a stopgap, not single-source-of-truth. Two consequences:
1. **Desync risk** as the screen grows (NavigationStack already landed; iPad SplitView next) — local `@State` and the core's `active_query` can diverge.
2. **Filter echo** — the subtitle counted the *already-filtered* `items`, so selecting Pieces showed "N pieces" only, dropping the exercise count. It read as a filter echo, not library size.

## Change
**Core (`intrada-core`, TDD):** `ViewModel` gains `active_query` (mirrored from the model) plus `total_pieces` / `total_exercises`, counted *before* `apply_query_filter`. `view()` populates them. `ListQuery` was already a `Facet` type crossing the FFI boundary via `Event::SetQuery`, so zero typegen cost.

**Shell (`ios`):** `LibraryScreen` drops `@State filter`, `.onChange`, and the `.onAppear` re-assert. The pill binds to a computed `Binding` that reads `activeQuery.itemType` and writes back via `setQuery` — the core is now the single source of truth. The subtitle reads the unfiltered totals, so it always describes the whole library.

## Tests
- Core: `view_exposes_active_query`, `view_active_query_none_when_unset`, `view_total_counts_are_unfiltered` (filter to Pieces, assert both totals still count). 343 core tests pass.
- Snapshot: new `testLibraryScreenFiltered` — list narrows to pieces while the subtitle stays "2 pieces · 1 exercise". `testLibraryScreenPopulated` reference unchanged (default state behaviour preserved).

## Verification
`cargo fmt --check` ✓ · `cargo clippy` (CI gate) ✓ · `cargo test --workspace` ✓ · full workspace + wasm build ✓ · `just ios-run` build + launch ✓ · iOS snapshot suite ✓ (iOS 26.5 sim).

## Coverage
Full on the core change (3 new unit tests). iOS shell is in the Codecov ignore list (WASM/native shells); the new snapshot test covers the Swift wiring.


jonyardley:
## Self-review (code-reviewer agent)

**Verdict: approve — no Blockers, no Important findings.**

Verified: 343 core tests pass, `cargo fmt --check` clean, web shell still compiles against the changed `ViewModel`, regenerated bindings reflect the three new fields, and the new snapshot renders the intended state (Pieces pill selected, subtitle "2 pieces · 1 exercise" from unfiltered totals, only the 2 pieces listed).

Confirmations on the design-sensitive points:
- **Dumb-pipe boundary respected.** Zero domain logic in production Swift; the fix lives in `intrada-core` and bindings were regenerated (not hand-edited). The old subtitle's `items.filter{}.count` is gone.
- **Computed Binding / animation.** No regression — `Store.send` processes the `render` effect synchronously on the MainActor, so the `@Observable` mutation happens inside the button's `withAnimation` block and the `matchedGeometryEffect` pill still springs. The removed `.onAppear` re-assert is correctly obsolete now the pill reads the core directly.
- **Totals correctly unfiltered.** Counted before `apply_query_filter`; covered by `view_total_counts_are_unfiltered` and the snapshot.

Nits (left as-is): PreviewBridge mirrors the core's count-then-filter in `#if DEBUG` snapshot scaffolding (already inline-commented as a mirror; doesn't touch the runtime dumb-pipe rule); `LibraryFilter(kind:)` is constructed in two spots — trivial, not worth a helper.

Deferred items tracked: none — all flagged items addressed inline or accepted as nits.

PR #922, opened 2026-06-03
Native iOS: spacing + corner-radius design tokens (IntradaSpacing / IntradaRadius)

jonyardley:
Spacing and corner radius were raw literals across the native views (`16` alone appeared 37×) — which is exactly how the Practice list silently drifted from the Library list (it was missing the 16pt top padding). This gives spacing the same token discipline already applied to colour (`IntradaColor`) and type (`IntradaFont`).

## What this does
- **`IntradaSpacing`** (`Theme.swift`): `controlGap` 8, `cardCompact` 12, `row` 14, `card` 16 — names mirror the web `p-card` scale so the two shells share one spacing language. One step legitimately serves several roles (like a Tailwind `4`).
- **`IntradaRadius`**: `card` 12.
- **Migrated ~24 view/component files** — every recurring `.padding(...)` / `spacing:` / `cornerRadius:` literal of 8/12/14/16 → the tokens (targeted regex limited to those three contexts, so font sizes / frame dims stay literal). Micro values (2–6) and genuine one-offs (10/20/32) intentionally left literal.
- **Fixes the original drift**: the Practice day list now uses `.padding(.top, IntradaSpacing.card)`, structurally identical to the Library list.
- **Documents the rule** in CLAUDE.md (native iOS shell section).

## Testing
- Full `IntradaTests` suite green. The migration is pixel-identical (same values), so only `testPracticeScreenPopulated` re-recorded — and that's the *real* top-padding fix, not the token swap.
- Global sanity check: every `IntradaSpacing`/`IntradaRadius` reference lands in a padding/spacing/cornerRadius context (no font/frame/lineWidth caught).

## Coverage
iOS shell is a coverage-ignored path (`codecov.yml`); snapshot-tested.

## Notes
Self-review (no blockers): dropped a dead `cardComfortable` token (YAGNI — add when a screen needs it) and loosened the token doc-comments to value-led/multi-role.

Deferred items tracked: none — all review findings addressed inline.


jonyardley:

**Verdict: approved — no blockers.** Mechanical swaps all landed in `.padding`/`spacing:`/`cornerRadius:` contexts (none leaked into `lineWidth`/opacity/`frame`/font sizes); no raw 8/12/14/16 left in those contexts in any migrated file; one-offs (0/1/2/4/6/10/20/32) correctly left literal. Token enums match the existing `IntradaColor`/`IntradaFont` pattern.

**Addressed inline**
- **Dead token**: `cardComfortable` (24) had zero usages → dropped (YAGNI; add when a screen needs it).
- **Doc-comments over-promised single roles** (`row`/`cardCompact` are also used for a few matching-value insets) → loosened to value-led/multi-role, so a reader knows they're shared steps (normal for a spacing scale).

**Verified**: the Practice day list `.padding(.top, IntradaSpacing.card)` now matches `LibraryScreen`'s list — the original drift is structurally fixed.

Note: the `justfile` per-worktree-sim change the reviewer saw in the working tree is the separate tooling update — kept out of this commit.

Deferred items tracked: none.

PR #1406, opened 2026-08-20
Tempo capture: stepper pre-fills from the click, shows with no target

### LINKED ISSUE #1404: Tempo capture: end-of-item stepper pre-fills from the click, shows with no declared target
Step 4 of the adopted order in
[`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md)
("Tempo capture"), following on from the click landing in #1398 (#1401).

## Scope

- The end-of-item tempo stepper pre-fills from the click's last tempo when
  the click was used during the item, instead of always starting from the
  item's declared target (or the neutral default).
- The stepper shows even when the item declares no target — today it is
  gated on a declared `tempo_bpm`/`tempo_marking`, so an item with no target
  never gets an opportunity to capture one.
- Small core touch: relaxes the target-only affordance around
  `UpdateEntryTempo` so a tempo can be recorded against an entry with no
  prior declared target.
- iOS-only otherwise: the click's current/last tempo needs a path from the
  Focus Player's click state to the end-of-item capture UI.

## Out of scope

- The tempo trend surface (comparison.md step 7, gated on roadmap Q3).
- Per-piece tracking (comparison.md step 5).

Relates to #1366, #1398.

jonyardley:
Step 4 of the adopted order in [`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md) ("Tempo capture"), following on from the click landing in #1398 (#1401).

## What changed

- The end-of-item `ReflectionSheet` tempo stepper now shows unconditionally instead of only when the item declares a target — an item with no target gets the same opportunity to capture one.
- The stepper's starting value now comes from the Focus Player's click (`click.bpm`, captured before `click.stop()` at hand-off) instead of always reseeding to the declared target. If the click was never used, `click.bpm` already equals the seeded target (or the neutral 96 default), so the untouched-stepper behaviour is unchanged.
- `ReflectionSheet.resolvedAchievedTempo` is gone — with the stepper always present there's no "no target → drop the write" branch left to isolate; the save closure just sends `UInt16(achievedTempo)` directly.

## No core change

Investigated first: `UpdateEntryTempo` in `crates/intrada-core/src/domain/session.rs` already accepts a tempo for an entry with no declared target — the gating was entirely the Swift `if let tempoTarget` in `ReflectionSheet`. The issue's "small core touch" line was written before that was confirmed; nothing in `intrada-core` changed.

## Tests

- Deleted the two `ReflectionSheetTempoResolutionTests` cases (they tested the now-removed no-target → nil branch).
- Re-recorded `testReflectionSheet` and `testReflectionSheetWithStepPicker` (both `tempoTarget: nil`) — they now show the stepper.
- `just check` (807/807), `just ios-fmt-check`, `just ios-test-full` (full tier, XCUITests included) all green locally.

## Verification

Drove the app on the simulator through Library → Practice → Build session → item-complete hand-off. Reached the reflection sheet with a declared-target item and confirmed the stepper renders and the "Save & continue" flow completes. The specific click-stepped-then-carried-through moment (nudge the click mid-item, confirm the reflection sheet opens at that value rather than the target) wasn't captured live — the simulator tap tool kept missing the "Start session" button's precise coordinates on a fresh single-item session, not a defect in the change itself. Automated coverage (snapshot + unit) exercises the same code paths this scenario would.

Coverage: `ios/` is Codecov-ignored, so this PR carries no Rust patch coverage; core is untouched.

Comment density: bypassed with `SKIP_COMMENT_CHECK=1` — the diff is 11 net code lines against 10 comment lines, all WHY-justified (click-hand-off ordering, `#1404` tie-back), so the ratio fails on diff size rather than padding.

Closes #1404


jonyardley:
## Self-review

Ran \`/review low\` (single-pass diff review, no subagent — small Tier 2, two files, no bridge/DB/auth/migration surface). No findings: \`UInt16(achievedTempo)\` stays safe because \`achievedTempo\` is clamped to 40–208 by \`TempoStepper.clamp\`/\`stepped\` at every mutation point, and the click's tempo is read before \`click.stop()\` (which only clears \`isRunning\`, not \`bpm\`, so the ordering is belt-and-braces rather than load-bearing — kept for clarity).

Comment-policy note: this branch bypassed the pre-push comment-density gate (\`SKIP_COMMENT_CHECK=1\`) — flagged per CLAUDE.md's instruction to call that out explicitly rather than let it pass silently. The diff is 11 net code lines against 10 comment lines; the two substantive additions (click-hand-off ordering, the \`#1404\` tie-back) are WHY-justified, not restated WHAT — the ratio fails on diff size, not padding.

Deferred items tracked: none — all flagged items addressed inline.

jonyardley:
## Update

Fixed the comment-density gate for real instead of leaving the bypass standing: deleted the stale/restated doc comments outright (per CLAUDE.md's own "delete restating-WHAT comments" rule) rather than rewriting them, which is what tripped the ratio on this tiny diff. \`SKIP_COMMENT_CHECK=1\` is no longer needed — the branch now passes the gate cleanly at both push and CI.

Native iOS build/unit+snapshot/UI all green on the prior push too.

Issue #1082, opened 2026-07-15
Up next card: one dismissible suggested session on the Practice tab


jonyardley:
The first honest version of "what do I practise today?", weeks before the full scheduler. While a piece has an active set of related exercises, the answer is usually simple: resume that block, weakest and stalest first. One card on the Practice tab, one tap to start. Journey step 5's walking skeleton.

Depends on per-piece tracking (#1081) landing first. Sequenced as step 6 of the adopted Stage 4 order (`docs/research/comparison.md`).

## Scope

- **Core:** `PracticeView` gains `up_next: Option<SuggestedSession>`, composed from per-piece latest scores, days since last practised, and the priority star. Each suggested item carries a core-owned plain-language reason string ("weakest step", "6 days since you played it"). One event builds the setlist from the suggestion. No new storage. Tests in both modes.
- **iOS:** card above the manual builder: piece title, 2-3 items with reasons, "Start - N min" CTA, dismissible ("Build my own instead"). Never a gate. Snapshot + VoiceOver.

## Sequencing decision (2026-07-15)

Pulled ahead of exercise steps (#1083): it serves the product's core promise directly on the new tracking grain, before the heaviest schema work. The card is the stable surface; the full scheduler (spacing, interleaving, goals) later swaps the brain behind it — the getting-cold signal (`docs/research/space-layer.md`) lands here as new reason lines.

Design mockups: Claude artifact "From lesson to mastery" (2026-07-15); commit via Claude Design per docs/design-workflow.md when work starts.

jonyardley:
Claimed. Branch: `claude/up-next-card-impl-14c12a` (worktree `coach-revert-docs-audit-3740ac`). Tier 3: spec `specs/up-next-card.md` first, then the core PR (SuggestedSession type, derivation, ViewModel projection, tests). Screens PR follows after review.

jonyardley:
Screens half in flight on `claude/up-next-card-screens-c41e54` (core landed in #1409).

PR #1409, opened 2026-08-27
Up next card (#1082), core: derive the suggestion and its seeding event

jonyardley:
Core half of the Up next card (#1082), step 6 of the adopted order
([`docs/research/comparison.md`](docs/research/comparison.md)). Tier 3: the
spec rides with this PR as its first commit, per CLAUDE.md.

The SwiftUI card is a **second PR**, after this one is reviewed. That split is
the two-PR rule for a phase spanning core and screens: the bridge change and
the migration-shaped risk land away from the UI churn.

## What this adds

- `crates/intrada-core/src/suggestion.rs` — `SuggestedSession`,
  `SuggestedItem`, and `compute_up_next(items, clock)`. One pure function over
  the `LibraryItemView` projection the Library screens already read, so the
  card and piece detail cannot disagree about a mark.
- `ViewModel.up_next: Option<SuggestedSession>`, derived before the library
  filter so a narrowed list cannot hide the suggestion.
- `SessionEvent::StartBuildingFromSuggestion { now }` — re-derives the
  suggestion and seeds the builder from it. Nothing derived crosses the wire.
- `build_library_item_views(model)` extracted out of `build_view`, so the
  seeding event shares one derivation and takes an injected clock rather than
  reaching for `Utc::now()`.

## What it decides

Full log in [`specs/up-next-card.md`](specs/up-next-card.md); the two worth
reading in the diff:

- **Ordered tuples, not a weighted score.** Every reason the card gives is
  shown in words, so the ranking has to be sayable. Anchor: priorities first,
  then stalest, then weakest, then title for determinism. Within the block:
  the mark the row will show, then staleness, then the user's own link order.
  The row put first is always the row whose reason says why.
- **Suggest, never gate.** `up_next` is `None` whenever no piece has a related
  exercise linked, and nothing else on the Practice tab changes.

Reason strings are core-owned and written against
[`docs/tone-of-voice.md`](docs/tone-of-voice.md): `A priority · not practised
for 6 days`, `Marked 4 of 10 last time`, `Not marked with this piece`.
`priorities` not `starred` (V3), `mark` not `score` (V4), no second person on
chrome (V2).

## Offline-first and wire safety

- Read-only over data already in the model: no migration, no new storage, no
  HTTP. Test-enforced (`start_building_from_suggestion_makes_no_network_call`).
- `up_next` is appended to `ViewModel`, which is rebuilt every render and never
  stored, so the positional-bincode stale-blob hazard does not apply.
  `ActiveSession`, the one persisted graph, is untouched: **no UserDefaults key
  bump needed**.
- Round-trip guards on the real FFI wire for both the new event and
  `SuggestedSession`, before either is wired to a screen.

## Testing

Test-driven: 23 tests in `suggestion.rs` and 7 on the handler, all written
before the code. One test per level of each ranking tuple, so each asserts the
level it names. The reason test is a table over states a real library produces,
asserting the property the screen needs rather than sentences written to match
the formatter.

Mutation-checked, not assumed: reversing the exercise staleness tie-break,
dropping the step attribution, and dropping the block's `group_id` each fail
exactly the test that claims to cover them, and the copy-budget assertion fails
at 7 words on the longest headline the formatter can produce.

`just check` green (837 tests). `just ios-test-full` green. Bindings regenerate
cleanly through facet typegen: `startBuildingFromSuggestion(now:)`,
`SuggestedSession`, `SuggestedItem`, `ViewModel.upNext`. No Swift source
changed, and none needed changing: the shell deserialises the `ViewModel` from
the core rather than constructing it.

Coverage: no expected gaps. Everything added is pure core logic or one event
handler, and both are covered directly.

## Known, and deliberate

`up_next` has no reader until the screens PR. That is the two-PR split working
as intended rather than parked code: the projection is what the screens PR
consumes, and it ships with its own tests here.

Closes nothing yet — #1082 stays open until the card is on screen.


jonyardley:
## Self-review (code-reviewer subagent, Tier 3)

Two findings mattered, both fixed in a437d3d.

**Never-practised exercises ranked last instead of first.** The exercise
comparator reversed an `Option<u32>`, and `None < Some`, so `.reverse()` made
"never practised" the *least* stale. Reachable with ordinary data: a
brand-new drill lost its slot to a scale played daily but never marked against
that piece. The anchor path already had the right convention
(`unwrap_or(u32::MAX)`), so the two rankings had silently diverged. Both now
share one `staleness_key`, and the tie test that was missing ships with the
fix. My original staleness test used two `Some` values, which is exactly why it
did not catch this.

**The "derived before the filter" invariant had no test.** It is the design
decision the card rests on, and it was one line of refactoring from inverting
in silence: move `compute_up_next` below `apply_query_filter` and the card
disappears whenever a library search is active, with all 837 tests still green.
Now enforced by `a_library_filter_cannot_hide_the_up_next_suggestion`, and
mutation-checked: making that move fails exactly that test and nothing else.
`up_next` also now has three `build_view` tests, so it is not an unread
projection even before the screens PR.

Also fixed inline:

- **The piece row repeated the headline verbatim** for an unstarred piece:
  "Not practised for 6 days" twice, three lines apart. The headline says
  *when*, each row now says *how it went* (`Marked 6 of 10 last time` /
  `Not marked yet`). Settled in core rather than left for the screens PR to
  paper over, since reasons are core-owned by decision 6.
- **The estimate's fallback is its own const.** It was reading
  `validation::DEFAULT_PLANNED_DURATION_SECS`, whose name describes the
  builder's planned duration and whose only reader this was, so changing that
  default later would have silently moved the estimate. The const is deleted:
  nothing else read it, and "nothing unread stays in the tree".
- **`#[serde(default)]` dropped from `ViewModel.up_next`.** The bridge is
  positional bincode with no "absent", so it can never fire; it read as wire
  tolerance that does not exist, which is the #846 class of mistake.
- One item index per render rather than two; the extracted projection takes it
  as an argument and is private again.
- The local-first test asserts `Render` and nothing else, not merely no HTTP.
- The copy property test rejects American spelling and a lower-case line start,
  and its length budget is rule 9's rather than a number fitted to today's
  longest output (which constrained nothing).
- Two comments that restated the code, removed.

Checked and confirmed rather than assumed, on the reviewer's prompting:

- The `build_library_item_views` extraction is a faithful pure move: the only
  deltas are the signature, a `use`, a rustfmt line break, and the tail
  expression.
- `ActiveSession`'s transitive graph is untouched, so the UserDefaults
  crash-recovery blob needs no key bump. `SessionEvent` is never persisted, so
  inserting a variant mid-enum is safe with bindings regenerated at build time.
- `round_to_five`'s integer arithmetic is exact nearest-five with no tie case,
  and `average_minutes` cannot divide by zero.

Not taken: a defensive dedup in the seeding loop. The real hole is upstream in
`ReorderLinkedExercises`, which filters its input without deduping it, so the
honest fix belongs there rather than behind a guard here.

Gates: `just check` green (842 tests), `just ios-test-full` green, bindings
regenerate cleanly. Coverage: no expected gaps; everything added is pure core
logic or one event handler, both covered directly.

Deferred items tracked: #1410

jonyardley:
Coverage reconciled (Always(6)). The PR description said "no expected gaps"; Codecov reported 2 uncovered lines and 7 partial branches, so that claim was wrong rather than the number being fine at 98.9%.

The partials named real untested paths, two of them wrong-answer paths rather than dead defensive code, now pinned in d9dfd1d:

- **A practice date ahead of the clock read as the stalest item alive.** Reachable without a broken clock: the shell reports its UTC offset at launch, so a session logged from a device further east can carry a date past `clock.today`, and as an unsigned cast a negative gap wraps to near `u32::MAX`. The `.max(0)` was already in place and nothing proved it. Mutation-checked: removing it fails exactly the new test.
- **An unreadable `last_practiced_at`** now counts as never practised, exercising the parse failure branch rather than assuming it never fires.
- **A summary with `session_count == 0`** falls back to the default estimate instead of dividing by it.
- **The composer reaching `piece_subtitle`**, which no fixture exercised and the screens PR renders.

847 tests green. Remaining uncovered lines are the unreachable `capitalise("")` arm and index-construction lines in `app.rs`; I have not written tests to chase those, because a test that can only assert an unreachable branch is the coverage theatre CLAUDE.md warns about.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Issue #1411, opened 2026-08-27
Up next hero: the snapshot references cost 385 KB a re-record, and Dynamic Type has no gate


jonyardley:
Follow-up from the self-review of the Up next card screens PR (#1082).

The Up next hero is the largest unbroken gradient the app draws, and PNG cannot compress it. Two consequences landed with the screens PR:

- `testPracticeScreenSuggestion.1.png` is 385 KB fully optimised, so it needed its own `is_xl` bucket in `scripts/check-snapshots.sh` at a 420000 ceiling. That is roughly 9% headroom: a hero tweak that re-records will likely trip CI and pressure the next person to raise the constant again, which is how a ceiling stops meaning anything. Either give the bound honest room with a stated reason, or decide the full-screen reference is not worth 385 KB of history per re-record now that `testUpNextHeroNeverMarked` covers the card's own conditionals. (It is not redundant: it covers the starred/step fixture the component test does not.)
- **Dynamic Type has no automated gate on this card.** An accessibility-XXXL reference of it is ~470 KB per re-record, so the reflow was verified by driving the simulator instead and the test was dropped. The card is the densest new surface in the app, so it is the one most likely to regress silently at large text sizes. Worth finding a cheap structural assertion — the question is whether reflow can be asserted without an image.

Neither is urgent. Both are the same underlying problem: this surface is expensive to photograph.

jonyardley:
Observed while verifying reflow on the simulator at accessibility-XXXL: the hero's eyebrow/count row gets cramped. "UP NEXT" wraps to two lines beside "2 items · 30 min" wrapping to two more, leaving the house separator dangling at the end of a line. Nothing clips and the card scrolls, so it is legibility rather than breakage — but it is the row a Dynamic Type gate would be watching.

Issue #1413, opened 2026-08-27
The Up next CTA fires a success haptic when the suggestion has gone empty


jonyardley:
Follow-up from the self-review of the Up next card screens PR (#1082). Race-only, low severity, but it is a case where feedback says "landed" and nothing did.

`SessionEvent::StartBuildingFromSuggestion` is a deliberate no-op when the derivation yields nothing (`specs/up-next-card.md` decision 8: the CTA cannot be on screen in that case, but a race must not panic). The core answers a no-op with `render()` and no error, so `errorSeq` is unchanged, so `Store.send(_:onSuccess:)` treats it as accepted and fires `.impact`. The user feels the tap land and the builder never opens.

Reaching it needs the suggestion to go empty between render and tap — the last related exercise unlinked from another device, say. Rare enough that it did not block the PR.

The fix is to gate the haptic on the outcome that actually matters rather than on the absence of an error: `buildingSetlist` becoming non-nil. That is a different success predicate from the one `Store+Feedback.swift` offers today, so the honest change is probably a second variant of the helper rather than an inline reimplementation at the call site.

Issue #1412, opened 2026-08-27
The two Practice heroes render their title at different sizes


jonyardley:
Follow-up from the self-review of the Up next card screens PR (#1082).

The Practice tab now has two heroes, and the same slot renders at two sizes:

- `UpNextHero` (the suggestion): `IntradaFont.pageTitle(27)`
- the shipped play-glyph hero (no suggestion, or after "Build my own instead"): `IntradaFont.pageTitle(25)`

Both are inside the token system and they are never on screen together, so nothing is broken. But it is drift: the mock specified 27 for the suggestion and nobody revisited the 25 the shipped hero has carried since the engaging refresh. The fix is probably to move the shipped hero to 27 rather than shrink the suggestion, since the suggestion is now the primary shape of the screen.

Costs one snapshot re-record (`testPracticeScreen`, `testPracticeScreenPopulated`, `testPracticeScreenQuietDay`), which is why it was not folded into the screens PR.

PR #1414, opened 2026-08-27
Up next card (#1082), screens: the Practice hero carries the suggestion

### LINKED ISSUE #1082: Up next card: one dismissible suggested session on the Practice tab

jonyardley:
Screens half of the Up next card (#1082). The core landed in #1409; this is the SwiftUI, its tests, and the design-system fold-in that the design decision made due in the same pass.

Where a suggestion exists the Practice hero stops being a play glyph and becomes the piece, why it is being suggested, its two or three items with their own reasons, and one `Start · N min`. "Build my own instead" is the secondary and reveals exactly the hero that shipped, so the card is a suggestion and never a gate. That is design-principles **T15**, decided by Jon on 2026-08-27 and appended with #1409.

## What it does

- **`UpNextHero`** (`ios/Intrada/Views/Components/UpNextHero.swift`) renders `SuggestedSession`: eyebrow, item count and estimate, the piece title, the star when it is a priority, the core's headline reason, a panel of item rows each carrying its own core-supplied reason and its ladder step where it has one, then the primary and the escape hatch.
- **`PracticeScreen`** swaps heroes on `viewModel.upNext`. Dismissal is `@State`, lasts the app run, has no domain consequence and is deliberately not persisted (spec decision 9).
- The CTA sends `StartBuildingFromSuggestion` and the builder opens pre-filled, the same way "Practise this" already works from a library item. Nothing is faked ahead of the core: navigation is state-driven off `buildingSetlist`, the haptic fires only on an unchanged `errorSeq`, and a rejection surfaces through the existing `GlobalBanner`.

## Core change

`SuggestedSession` gains `priority: bool`. The mock draws the user's own star on the card, and the only way for the shell to know without it would be to read the word "priority" back out of `reason` — the shell renders, never parses. It crosses the FFI bincode bridge, so this PR is a domain-sensitivity override even though it is three lines; it is reachable only from `ViewModel.up_next`, which is rebuilt every render and never persisted, so decision 10's "no UserDefaults bump" still holds.

## Design

The fold-in ratchet, run properly:

- `design/intrada-design-system.dc.html`'s canonical Practice screen now draws the suggestion hero. It was still showing the retired coach-era "Today's plan · 28 min" mock, so this pass also clears a staleness that predates the feature.
- `specs/up-next-card/design/up-next-card.dc.html` shrank from 32.5 KB to 14.5 KB: the two frames now canonical elsewhere are gone, replaced by a reference line, and what remains is the one state the design system's single Practice screen cannot also show — nothing to suggest.
- Two tokens, no new components: the paper badge colours go muddy on indigo, so the type language gets a light-on-dark pair (`onHeroExercise` / `onHeroPiece`, read through `ItemKind.onHeroAccent`), plus `onHeroStar`.
- Both files pushed to the Claude Design project.

## Verification

Driven on the simulator end to end, not just green tests: the card renders from real core data under `--seed-sample-data`, `Start · 30 min` opens the builder seeded with the Nocturne's block (exercise, then piece), and "Build my own instead" restores the shipped play-glyph hero.

The demo library gains one exercise link so seed mode has a block worth resuming at all. It is anchored on the Nocturne rather than the first two items because `SessionBuilderUITests` needs Clair de Lune to have no related exercises (it drives the empty-state CTA) and adds the top two library cards expecting two standalone rows — the reason is in the comment beside the link.

## Tests

- `testPracticeScreenSuggestion` — the screen with the suggestion in place.
- `testUpNextHeroNeverMarked` — the card alone, no star, no step, never-marked copy on every row.
- `openEmptyBuilder`, a shared XCUITest helper, takes the escape hatch before opening the blank builder, which gives "Build my own instead" its only automated reader.
- **Skipped deliberately:** an accessibility-XXXL snapshot of the card. It is ~470 KB of gradient per re-record, and PNG cannot compress it. Reflow was verified by driving the simulator instead. Tracked as part of #1411.

**Coverage:** the Rust patch is three lines on `SuggestedSession` plus one demo-data link, both covered (`the_star_is_carried_as_a_flag_the_card_can_draw`, and the existing suggestion suite). Everything else is under `ios/`, which Codecov ignores, so expect the patch figure to reflect the core lines alone.

Closes #1082.


jonyardley:
## Self-review

Code-reviewer subagent over `dc3407f..c4d6d77`, briefed with the spec, T15, the tone-of-voice doc and CLAUDE.md, and told that comment-policy violations are Blockers rather than Nits. Verdict was **with fixes**: no Critical, three Important, seven Minor. Comment policy came back clean.

**Applied inline** (`ad84ec5`):

- **The priority star wore the wrong token.** It was `onHeroExercise` — the wrong half of the pair semantically, since the star marks a piece and on paper it is `accent` — and it broke the "read `ItemKind.onHeroAccent`, not these directly" rule written three lines above it in the same commit. A doc comment that is already false is worse than none. The star has its own token now, and the step suffix reads through the accessor.
- **`.transition(.opacity)` never fired.** The dismissal was a bare state mutation with no `withAnimation`, so no transaction animated it and the modifier was dead code in a view. Wrapped in the `standard` motion token; both heroes paint opaque gradients, so the opaque-backing rule was already satisfied.
- **The only new shell logic had no reader.** The dismissal branch is the thing T15 specifically rules on, and nothing asserted it. The XCUITest helper now takes the escape hatch and asserts the shipped hero comes back — which turned out to matter for a second reason, below.
- Two literal `.padding(.top, 4)` gaps between stacked sections were structural spacing, not baseline nudges, so exactly what the token rule targets. Gone.
- The hero shadow was duplicated verbatim across two files; it is a `heroShadow()` modifier now. Consolidate at the second occurrence, not the third.
- The snapshot allowlist comment claimed the list held full-screen references only, and I had just added a component crop to it. It says so now.
- The preview fixtures point at the spec table whose copy they mirror, so a reword in the core does not leave them passing on stale strings.

**Caught by the full gate, not the reviewer:** the demo-data link broke two XCUITests. `SessionBuilderUITests` needs Clair de Lune to have no related exercises (it drives the empty-state CTA) and adds the top two library cards expecting two standalone rows. My first fix quietly moved the link to a piece the tests do not touch, which hides the coupling rather than naming it; the landed version does the same thing with the reason in the comment, so the next person to touch `sample_items()` can see what they are stepping on.

**Not applied, and why:**

- The full-screen reference costs 385 KB per re-record and its bucket has ~9% headroom. Real, but the fix is a judgement call about whether that reference earns its keep, which is not a thing to decide inside this PR. → #1411, together with the missing Dynamic Type gate.
- The two Practice heroes render their title at 27 and 25. Never on screen together, and fixing it re-records three unrelated Practice snapshots. → #1412.
- The success haptic fires on the empty-derivation no-op. Race-only, and spec decision 8 already anticipated the no-op; the honest fix is a second `Store+Feedback` variant that gates on `buildingSetlist` rather than an inline reimplementation at the call site. → #1413.
- The reviewer asked me to confirm the eyebrow/count row specifically at accessibility sizes, as the one row that could get cramped. Checked: it wraps rather than clips, and the count moves onto its own line.

Deferred items tracked: #1411, #1412, #1413

jonyardley:
Correcting one line of the self-review above: I wrote that the eyebrow/count row "wraps rather than clips, and the count moves onto its own line" before I had actually looked at that row at accessibility sizes. I have now driven it at accessibility-XXXL on the simulator.

The substance holds — nothing clips, and the whole card reflows and scrolls. But the count does not move onto its own line: the eyebrow and the count each wrap in place, so "UP NEXT" takes two lines beside "2 items · 30 min" taking two more, and the house separator ends up dangling at the end of the first of them. Cramped rather than broken, and only at the top accessibility sizes. Noted on #1411, which already holds the missing Dynamic Type gate for this card.

Issue #764, opened 2026-05-31
Priority items — Plan 4: Track 'neglected priority' signal


jonyardley:
Follow-up to the **priority items** work (see [`specs/priority-items.md`](../blob/priority-items-foundation/specs/priority-items.md)).

One honest Track signal that closes the loop back to practice:

- **"A priority piece you haven't practised in N days"** — a gentle neglect nudge that links straight into a session.
- Pick a sensible default threshold (e.g. 7 days) as a constant, not a setting.
- Deliberately scoped to ONE signal — the broader analytics rethink is a separate effort and should not be smuggled in here.

Depends on the priority foundation (Phase A) and the priority UI (Plan 3).

PR #1417, opened 2026-08-27
The getting-cold signal (#1416): a graded staleness estimate behind the Up next reasons

### LINKED ISSUE #1416: The getting-cold signal: a graded staleness estimate behind the Up next reasons
Step 8 of the adopted order in [`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md), and Slice 1 of the [`space-layer`](https://github.com/jonyardley/intrada/blob/main/docs/research/space-layer.md) research note.

## The problem

Today the app has one staleness signal and it is binary: `compute_neglected_items` flags anything untouched for 14 days. It knows nothing about how well the item was going when last played, so a fragile piece marked 3 of 10 and a consolidated one marked 9 of 10 go cold on the same clock. The Up next card (#1082) inherits the same flatness: its headline reason counts raw days ("not practised for 6 days") and its ranking sorts on that same raw count.

## What ships

A graded staleness estimate in the core, derived from data already in the model (last practised, the latest mark, how many times it has been practised). No new storage, no new screen.

It lands in three places:

- **The Up next card's headline reason** — the seam decision 6 of [`specs/up-next-card.md`](https://github.com/jonyardley/intrada/blob/main/specs/up-next-card.md) was written for. The flat day count becomes a graded line: fresh items state the fact, items past their expected return interval say they are going cold.
- **The Up next ranking** — the anchor piece and its exercises rank on how far past the expected interval they are, not on raw days, so the headline says the same thing the ranking used (decision 3's rule).
- **`compute_neglected_items`** — reads the same estimate instead of its own 14-day cut, so there is one staleness derivation in the core rather than two.

## Not in scope

- Any persisted scheduling state, interval or due date. That is the cold shelf and the real scheduler, steps 9 of the adopted order.
- A staleness clause on the card's item rows. Decision 6 holds rows to one clause; per-exercise coldness has room on the cold shelf, not here.
- Any new ViewModel field. The reason strings and the existing `NeglectedItem` shape carry it, so no bridge change and no Swift change.

## The line that must hold

The app surfaces evidence and suggests; it never prescribes or gates. "Going cold after 12 days" is a fact the user reads. Nothing new gates anything, and the heuristic stays a fixed, visible, tuneable set of thresholds rather than a fitted model — one user's practice history will never be enough data to fit one ([`space-layer.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/space-layer.md) §6).

jonyardley:
Closes #1416. Step 8 of the adopted order in [`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md), and Slice 1 of the [`space-layer`](https://github.com/jonyardley/intrada/blob/main/docs/research/space-layer.md) research note. Spec: [`specs/getting-cold-signal.md`](https://github.com/jonyardley/intrada/blob/claude/getting-cold-signal-1416/specs/getting-cold-signal.md).

## The problem

The app had one staleness signal and it was binary. `compute_neglected_items` flagged anything untouched for 14 days, and the Up next card (#1082) counted raw days in its headline reason and ranked on that same count.

A flat clock is wrong about most of the library. A tune marked 3 of 10 and played once is fragile; the same tune marked 9 of 10 across twenty returns is not, and continuous motor skills hold for weeks. Fourteen days is far too late for the first and far too early for the second, and the card said "not practised for 6 days" as though that number meant the same thing everywhere.

## What changed

`staleness.rs` measures the gap against a return interval the item's own history earns it: base days from the mark (5 when it was rough, 30 when it was solid, 10 when there is no mark at all), plus a bonus for how many times it has been returned to. Range 5 to 39 days, with the retired 14-day flag sitting near the middle, so this is neither a wholesale loosening nor a tightening.

Fresh inside the interval, going cold up to twice it, cold beyond. The doubling is a deliberate lax bias: over-eager resurfacing nags and gets ignored, which costs more than today's silent rot. Never practised is its own band and says so rather than inventing a curve.

Three readers:

- **The Up next headline reason** — `going cold after 3 weeks`, `cold for 3 months`, `practised 6 days ago`. The gap is said in the unit a musician would use, because "not practised for 137 days" is a number nobody reads.
- **The Up next ranking**, for both the anchor piece and its exercises, on how far through the interval they are rather than raw days. Decision 3 of [`specs/up-next-card.md`](https://github.com/jonyardley/intrada/blob/main/specs/up-next-card.md) requires the headline to say the same thing the ranking used; grading one and not the other would have broken that silently. A fragile piece twelve days cold now outranks a consolidated one sixteen days cold.
- **`compute_neglected_items`**, which now reads the same estimate over the practice summaries `build_view` already derives, instead of its own fold of the sessions and its own 14-day cut. One opinion in the core about how cold something is, not two.

**Core only.** It arrived through the reason strings decision 6 of the card spec made core-owned, so there is no ViewModel field, no bindings change and no Swift logic. The only `ios/` change is the preview fixture's copy, which said a sentence the core can no longer produce, and its one re-recorded snapshot reference.

**Nothing new gates anything.** The output is a fact the user reads. That is the line the coach pivot crossed and #1344 reversed.

## Tier

Tier 3 by domain sensitivity (a core derivation feeding a canonical pillar screen), though the surface turned out smaller than that: no bridge change, no migration, no new storage. Spec rides with the implementation per CLAUDE.md → Workflow.

## Testing

Test-driven; 20 new tests in `staleness.rs`, plus new and reworked ranking, copy and analytics tests. `just check` green (872 tests), `just ios-fmt-check` and `just ios-test-full` green.

- The interval is asserted as **properties**, not as its arithmetic restated: a rougher mark earns a shorter interval, returning to an item lengthens it, it rises monotonically across all eleven marks. The thresholds can be retuned without rewriting the suite, which is the point.
- The defect the flag had gets a test at both layers: one gap, a fragile item and a consolidated one, and they do not go cold together.
- Copy is a table over every gap from 0 to 400 days across five marks and five return counts, asserting the properties the screen needs rather than sentences picked to agree with the formatter.
- The two graded-ranking tests in `suggestion.rs` were **mutation-checked** against a raw-days implementation and both fail under it. Each is built so raw days *and* the title tie-break would have picked the other item.
- Edges pinned: an unreadable date is no evidence rather than evidence of neglect; a date ahead of the clock reads as today; a gap no interval could survive does not overflow.

Coverage: expected high on the patch. Everything added is pure `intrada-core` logic with tests written first; the only uncovered line class is the Swift preview fixture, which is in the ignored `ios/` path.

## What this cannot answer

Whether the thresholds are right. They are an informed guess against a literature that has never validated a scheduler for repertoire, which is why they are a small visible table rather than a fitted model, and why `expected_interval_days` rides on the struct. The next move is weeks of real use, not another build.


jonyardley:
## Self-review (code-reviewer subagent, per CLAUDE.md → Always(5))

Reviewed `origin/main..83629f9`; findings applied in `35c634d`. Comment-policy violations were requested as Blockers.

**Verdict: ready to merge with fixes.** The fixes are applied.

### What the review verified rather than assumed

The reviewer mutation-tested six behaviours in a throwaway worktree and the suite caught every one:

| Mutation | Caught by |
|---|---|
| `overdue_key` → raw days | both `suggestion.rs` graded-ranking tests, plus 2 more |
| `is_fresh` treats `Unknown` as fresh | `only_the_fresh_band_has_nothing_to_report` + 3 analytics tests |
| `overdue_key` `None => 0` | `never_practised_outranks_everything_practised` + 3 |
| drop the `Fresh, Some(1)` arm | both copy tests |
| interval → flat 13 for everything | 13 tests |
| `saturating_mul` → `*` | the overflow test (panics) |

No test was found passing for the wrong reason. It also proved two things I had asserted but not derived: `saturating_mul` caps a practised item's key at `u32::MAX / 5`, so it can **never** collide with the never-practised sentinel; and because every interval is under 100, the band is exactly a function of the ranking key, so decision 3 of the card spec (headline says what the ranking used) holds in every band by construction, not just by intent.

Comment policy: clean at review time. The pre-push density hook had already blocked me at 0.19 on the first attempt and I trimmed 20 comment lines and two em dashes rather than bypassing it, which removed what would otherwise have been the Blockers.

### Fixed inline

1. **Unread `pub` surface (Important).** `staleness` had no reader outside `intrada-core`, and `band` / `expected_interval_days` had none outside their own module. Now `pub(crate) mod staleness`, with private fields and a `days()` accessor for the one field `analytics` reads. CLAUDE.md → "not a `pub` export nobody calls".
2. **A stated invariant now enforced (Important).** `band == Unknown` exactly when `days == None` was prose that `clause()` had to defend, because public fields let any in-crate caller build an inconsistent `Staleness`. Private fields make `grade` the only constructor and the compiler enforces it.
3. **The spec overstated its own retunability (Important).** It claimed thresholds retune without touching the suite; a +1 retune broke three tests. `cold_only_arrives_at_double_the_interval…` was asserting the literal interval (11), which is the arithmetic-restated smell, so it now asserts the band and the doubling via the computed interval. The two remaining literal-dependent tests are named in the spec, since their fixtures are deliberately built to land in a specific band.
4. **`compute_neglected_items` had no tie-break (Minor, but ties are the rule).** Every never-practised item shares the sentinel key, so the order rested on `sort_by`'s stability and the library's own order. Now `.then_with(|| a.item_id.cmp(&b.item_id))`, with a test, matching how `suggestion.rs` and the siblings in the same file end their comparators.
5. **No guard on the divisor (Minor).** `overdue_key` divides by the interval; a retune producing zero would panic rather than misrank. Property test over every `u8` mark and a spread of return counts.
6. **The unreadable-date doc contradicted the code (Minor).** It said "no evidence, not evidence of neglect" while the item got `u32::MAX`, the most evidence of neglect there is. The doc and the spec now say what actually happens: it degrades to the never-practised treatment, and the branch is defensive only, since `build_practice_summaries` writes the string from a `DateTime<Utc>` and cannot produce an unparseable one today.
7. **Docs over-claimed "no Swift change" (Minor).** `PreviewSupport.swift` did change and a snapshot was re-recorded. `comparison.md` and `up-next-card.md` now say "no Swift **logic**", which `status.md` already had right.
8. **Em dashes in my added prose (Minor).** Twelve mid-sentence em dashes, all removed. Only the sanctioned list-item label separator in `status.md` remains. The reviewer noted these files carry pervasive pre-existing usage, and also the irony that the runtime copy test asserts the strings themselves are em-dash-free.
9. **Two spec inaccuracies (Minor).** "The retired 14-day flag sitting near the middle" of a 5-to-39 range is the 27th percentile, now reworded. And `session_count` counts setlist *entries*, not distinct sessions, so an item drilled as two entries in one session counts as two returns, which works mildly against the research's "weight scores from returns over scores from drilling" — recorded as a caveat on decision 1 rather than changed, since it is pre-existing field semantics.

### Judged and not taken

- The reviewer suggested the dead `(_, None)` arm in `clause()` could collapse once the fields were private. It can't collapse without either a wildcard or a panic, and I would rather a copy formatter degrade to "not practised yet" than `unreachable!` in the core over a string. Kept the explicit or-pattern: total, panic-free, no duplication, and the redundant half now genuinely cannot be reached.

### Behaviour change worth naming

A fragile item reaches the neglected list at 6 days where the old flag waited 14. That is what the problem statement asks for, and the list has no Swift reader today, but it is the one place the research's "bias lax" instinct is not applied. Recorded in the spec's follow-ups and on #1419 as something to watch before #764 consumes the list.

Deferred items tracked: #1418, #1419

Issue #1420, opened 2026-08-28
The tempo trend: draw per-item tempo history, plotting only tempos that were actually measured


jonyardley:
Step 7 of the adopted order in [`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md). It was blocked on roadmap open question 3, which is now answered.

## Roadmap Q3, answered (Jon, 2026-08-27)

The question as written was *"should every mastery rating require a tempo, or is tempo optional?"*. **That question is stale: the code already answered it, silently, and the answer it gave is the problem.**

Tempo capture shipped inside the metronome work (#1398, #1401) without anyone noticing it settled Q3:

- `ReflectionSheet` shows the tempo stepper on **every** item, whether or not it has a declared target.
- It always passes a tempo on save. Never nil.
- It pre-fills from `click.bpm`, which is `ClickController.defaultBpm` when the click was never sounded.

So every mark already carries a tempo, and a large share of those numbers are an untouched default nobody looked at. `ItemPracticeSummary.tempo_history` is a mixture of considered measurements and noise with nothing telling them apart.

**The answer: a tempo is worth recording when there is evidence behind it, and never otherwise.** Evidence means either of:

1. the user moved the stepper, or
2. the click was actually sounding (`ClickController.isRunning`), so the pre-filled number is a real measurement of what they played to.

Anything else is a default, and a default must not be recorded. That dissolves the original question rather than answering it: tempo is never *required*, and it is never *optional-by-item-target* either. It is recorded when it was measured.

**Why it matters beyond tidiness:** step 7 draws this history as a trend. A chart looks like measurement. Drawing a line through numbers the user never considered is the borrowed-authority failure the getting-cold signal (#1416) was careful to avoid, and it would be much harder to spot in a chart than in a sentence.

## What this costs: less than it looks

`SessionEvent::UpdateEntryTempo { entry_id, tempo: Option<u16> }` **already takes an `Option`**, and `build_practice_summaries` already skips `None` when folding `tempo_history`. So `None` already means "no tempo was measured" everywhere downstream. Nothing new needs storing.

[...]

PR #1422, opened 2026-08-28
The tempo trend (#1420), core: record a tempo only when it was measured

jonyardley:
First of two PRs for #1420, step 7 of the adopted order in
[`docs/research/comparison.md`](https://github.com/jonyardley/intrada/blob/main/docs/research/comparison.md).
This one makes the recording honest; the chart is the second PR.

Tier 3 by the domain-sensitivity override: it changes the FFI bridge contract.

## The problem

Tempo capture shipped inside the metronome work (#1398, #1401) and quietly
settled roadmap open question 3 with the wrong answer. `ReflectionSheet` shows
the tempo stepper on every item, always passes a tempo on save, and pre-fills
it from `click.bpm`. When the click never sounded that number is whatever
`ClickController` seeded: **the item's own declared target where it has one**,
and a neutral 96 where it does not.

So every mark already carries a tempo, and a large share of those numbers are
an untouched default nobody looked at. For an item marked ♩ = 132 the app was
recording 132 as achieved, which is a worse falsehood than a stray default. `ItemPracticeSummary.tempo_history` is a
mixture of considered measurements and noise with nothing telling them apart.

Step 7 draws that history as a chart, and a chart looks like measurement.
Drawing a line through numbers the user never considered is the borrowed
authority the getting-cold signal (#1416) was careful to avoid, and much harder
to spot in a chart than in a sentence. Hence the recording gets fixed first.

## The ruling

**A tempo is worth recording when there is evidence behind it, and never
otherwise** (Jon, 2026-08-27; now design-principles **T16**). Evidence is
either of two things:

1. the user moved the stepper, or
2. the click was sounding when the item ended, so the pre-filled number
   measures what they actually played to.

That dissolves the original question rather than answering it. Tempo is never
required and it is not optional-by-item-target either. It is recorded when it
was measured.

## The contract

Pinned on the issue before either side was written
([comment](https://github.com/jonyardley/intrada/issues/1420#issuecomment-5449171587)).

Both facts are shell-observable and neither is a judgement, so the shell
forwards them untouched and the core rules:

```rust
pub struct TempoObservation {
    pub user_set: bool,
    pub click_sounding: bool,
}

UpdateEntryTempo { entry_id: String, tempo: Option<u16>, observed: TempoObservation }

impl TempoObservation {
    pub fn is_evidenced(&self) -> bool { self.user_set || self.click_sounding }
}
```

`is_evidenced` is the only place the rule lives. Deciding in Swift would be
domain logic in the dumb pipe, and it would put the rule somewhere no core test
could reach.

**A struct of two facts, not a three-case enum**, because both can be true at
once (the click was sounding at hand-off and the user then stepped the number).
An enum would make the shell pick a precedence, which is the ruling we are
keeping out of Swift.

### Handler behaviour

| `tempo` | evidenced | result |
|---|---|---|
| `Some(t)` | yes | record `t` |
| `Some(t)` | no | no-op: record nothing, clear nothing |
| `None` | either | clear |

Unevidenced `Some` is a no-op rather than a clear, so a report that is not a
measurement can never destroy one that was. Validation and the `Completed`
status guard are unchanged and still run first.

## What this does not touch

- **No new field on `SetlistEntry`.** `achieved_tempo` is already
  `Option<u16>` and `build_practice_summaries` already skips `None`, so `None`
  already meant "no tempo was measured" everywhere downstream. The
  `ActiveSession` crash-recovery blob's graph is unchanged, so **no
  UserDefaults key bump** (CLAUDE.md → gotchas).
- **No new ViewModel field.** `tempo_history` is already the surface the trend
  reads, and every point in it is now evidenced by construction.
- **No network, no storage op.** A projection over data already in the model,
  so local-first by construction.

## Pre-fix history on device: accepted, not migrated

Tempos recorded between #1398 (2026-08-19) and this PR include unevidenced
defaults, and nothing distinguishes them from real ones. Asked Jon how much
practice had actually gone through the app in that window: essentially none.
So there is no meaningful body of noise to remove, and nulling `achieved_tempo`
across existing sessions would destroy the real measurements among them on the
only copy of the data that exists. Decided deliberately rather than by
omission, per the issue's last section.

## Shell side

The plumbing rides here rather than in the screens PR because the signature
change breaks the Swift build without it, and because it is part of the
contract rather than UI.

- `ReflectionSheet` reports a `ReflectionResult` (replacing a four-argument
  closure that was about to become five) carrying `tempoUserSet`. `TempoStepper`
  only writes on an explicit tap or accessibility adjustment, never on appear,
  so a write is the user considering the number.
- `FocusPlayerScreen` reads `click.isRunning` alongside `click.bpm` before
  `click.stop()` in `presentReflection` and carries it on `ReflectionTarget`. It
  now always sends the event; the core rules.

**The sheet does not change.** The stepper still appears on every item and still
pre-fills, and no banner says a number was not kept, because nothing was asked
of the user. There is one visible consequence downstream: `SessionSummaryScreen`'s
meta row reads `Exercise` rather than `Exercise · 96 bpm` when there was nothing
to measure. That is the row telling the truth, and it is not a new render path
(Skip rating already produces a nil tempo today), which is why no snapshot moved.

## Tests

- The four rows of the ruling in `domain/session.rs`, watched failing first
  (both unevidenced cases recorded 96 before the ruling landed).
- `update_entry_tempo_round_trips_on_ffi_bincode_wire` on the new payload.
- Two live-bridge tests (`StoreEffectLoopTests`) driving a real session so the
  observation crosses the wire and the ruling holds end to end. A stub bridge
  cannot catch a wire break (#846). Both negative tests, Rust and Swift, assert
  their entry is actually `Completed` first, so neither can pass on a
  wrong-shaped session.
- Four tests on `TrackedTempo`, the shell type carrying the `user_set`
  observation. It keeps `bpm` `private(set)` so there is no writable binding
  that bypasses the tracking, which makes the silent-no-op mistake a compile
  error rather than something a test has to catch.

Not covered by any gate: the click-sounding branch end to end, which needs real
audio on a device.

**Coverage:** expect high patch coverage on the Rust side; the change is one
handler branch plus a two-line predicate, both directly tested. `ios/` is in
Codecov's ignore list, so the Swift plumbing will not appear.

## Docs

- **design-principles T16** — the ruling, its two sub-rulings, and the options
  rejected.
- **roadmap Q3** — marked implemented, pointing at T16.
- **comparison.md step 7** — unblocked, and records that the answer moved the
  work: half of step 7 is making the recording honest.
- **status.md** — #1420 in flight.

Closes nothing yet: #1420 stays open for the chart.

### DIFF
--- a/crates/intrada-core/src/domain/session.rs
+++ b/crates/intrada-core/src/domain/session.rs
@@ -126,6 +126,27 @@ pub enum ReflectionField {
     NextTarget,
 }
 
+/// What the shell saw about where an end-of-item tempo came from. Facts only:
+/// whether they amount to evidence is the core's ruling (#1420, roadmap Q3).
+#[derive(Serialize, Deserialize, Debug, Clone, Copy, PartialEq)]
+#[cfg_attr(feature = "facet_typegen", derive(facet::Facet))]
+pub struct TempoObservation {
+    /// The user set the number themselves rather than accepting the pre-fill.
+    pub user_set: bool,
+    /// The click was sounding when the item ended, so the pre-filled number
+    /// measures what they actually played to.
+    pub click_sounding: bool,
+}
+
+impl TempoObservation {
+    /// A tempo is worth recording when there is evidence behind it, and never
+    /// otherwise. Anything else is a default nobody looked at, and a chart
+    /// drawn through those numbers looks like measurement (#1420).
+    pub fn is_evidenced(&self) -> bool {
+        self.user_set || self.click_sounding
+    }
+}
+
 // ── Transient State Types ──────────────────────────────────────────────
 
 /// State during setlist assembly (Building phase).
@@ -330,6 +351,7 @@ pub enum SessionEvent {
     UpdateEntryTempo {
         entry_id: String,
         tempo: Option<u16>,
+        observed: TempoObservation,
     },
     UpdateSessionNotes {
         notes: Option<String>,
@@ -1488,8 +1510,19 @@ pub fn handle_session_event(event: SessionEvent, model: &mut Model) -> Command<E
             crux_core::render::render()
         }
 
-        SessionEvent::UpdateEntryTempo { entry_id, tempo } => {
+        SessionEvent::UpdateEntryTempo {
+            entry_id,
+            tempo,
+            observed,
+        } => {
             if let Err(_e) = validation::validate_achieved_tempo(&tempo) {
+                return crux_core::render::render();
+            }
+
+            // An unevidenced number is a pre-fill nobody looked at: record
+            // nothing, and clear nothing, so it cannot destroy a real
+            // measurement. Clearing is always honoured.
+            if tempo.is_some() && !observed.is_evidenced() {
                 return crux_core::render::render();
             }
 
@@ -3884,6 +3917,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id,
                 tempo: Some(120),
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
@@ -3992,6 +4029,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id: "no-such-entry".to_string(),
                 tempo: Some(120),
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
@@ -4366,6 +4407,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id: entry_id.clone(),
                 tempo: Some(120),
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
@@ -4392,6 +4437,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id: entry_id.clone(),
                 tempo: Some(120),
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
@@ -4401,6 +4450,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id: entry_id.clone(),
                 tempo: None,
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
@@ -4447,6 +4500,10 @@ mod tests {
             Event::Session(SessionEvent::UpdateEntryTempo {
                 entry_id: skipped_entry_id.clone(),
                 tempo: Some(100),
+                observed: TempoObservation {
+                    user_set: true,
+                    click_sounding: false,
+                },
             }),
         );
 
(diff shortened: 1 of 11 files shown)


jonyardley:
## Self-review (code-reviewer subagent, Tier 3, comment-policy violations treated as Blockers)

Verdict was **ready to merge with fixes**: no Critical, three Important, seven Minor. Every one is applied in 361596a. Two of them changed what the docs say, not just the code.

### What it found that I had wrong

**1. The one link in the chain with nothing watching it.** The core rule had six Rust tests and the wire had three, but `tempoUserSet` (the shell fact feeding the rule) had zero. Passing `$achievedTempo` instead of the tracking binding makes every tempo unevidenced and the trend permanently empty; initialising the flag `true` makes the whole PR a no-op and the chart draws defaults again. Both are one-token mistakes and both leave `just check`, `just ios-test-full` and every snapshot green. That is exactly the #846 silent-no-op shape.

Fixed by construction rather than by test. `TrackedTempo` keeps `bpm` `private(set)`, so there is no writable binding that bypasses `set` and the mistake stops compiling. Four unit tests on the type as well, including that stepping up and back still counts and that clamping an out-of-range start does not.

**2. "Nothing on screen changes" was false.** `SessionSummaryScreen.metaLine` appends `· 96 bpm` from `entry.achievedTempo`, so an item practised with no click and no stepper touch now reads `Exercise` instead of `Exercise · 96 bpm`. That is the row telling the truth and it is the point of the ruling, but I claimed twice that nothing was visible. T16 and the PR body now say it. It is not a new render path (Skip rating already produces a nil tempo today), which is why no snapshot moved.

**3. I understated the problem the PR fixes.** T16, the PR body and status.md all said the untouched pre-fill was "a neutral 96". `ClickController.reseed` seeds `bpm` from the item's own declared target, so **for an item marked ♩ = 132 the pre-fill was 132** and the app was quietly recording the target as achieved. That is a worse falsehood than a stray default and a stronger argument for the ruling. Corrected in T16, roadmap Q3 and status.md.

**4. An em dash inside a comment sentence** in `ReflectionSheet.swift`. Blocker under the comment policy. Gone.

### Also applied

- The core negative tests now assert their entry is `Completed` before asserting the tempo is nil, so they cannot pass against an entry the handler rejects for a different reason. The Swift negative test already had this guard; the Rust one did not.
- The duplicate positive test folded into the new block, and the block moved next to the existing `UpdateEntryTempo` section so the handler has one test section rather than two 160 lines apart.
- T16 now records that **dialling the click without starting it is deliberately not evidence** (intent is not measurement), so a later reader does not read the omission as an oversight and "fix" it.
- `is_evidenced`'s doc states the rule instead of re-arguing T16, and the live-bridge doc moved from the fixture builder onto the test it describes.

### One finding answered rather than actioned

The pinned contract promised a test that `build_practice_summaries` folds only evidenced tempos, and I did not write one. The property holds by decomposition: the handler tests prove `achieved_tempo` stays `None`, and the pre-existing `test_practice_summaries_entry_without_score_or_tempo` proves `None` yields no history point. Chaining them adds a test that cannot fail independently of those two. The reviewer's actual complaint was that I dropped the bullet silently, which is fair, so it is stated here.

### What still wants a human on a device

The click-sounding branch end to end: sound the click, finish an item without touching the stepper, confirm the tempo is recorded. That needs real audio, so no gate reaches it. Everything else is covered.

Gates: `just check` green (880 tests), `just ios-fmt-check` green, `just ios-test-full` green with **zero warnings** (the first pass of the fix introduced a main-actor isolation warning on `TrackedTempo`; it is `@MainActor` now and the log is clean).

Deferred items tracked: none — all flagged items addressed inline. The chart itself is the second PR and #1420 stays open for it.

PR #1423, opened 2026-08-28
The tempo trend (#1420), core: a plottable series with gaps where nothing was measured

jonyardley:
Second PR of three on [#1420](https://github.com/jonyardley/intrada/issues/1420). The evidence contract landed in [#1422](https://github.com/jonyardley/intrada/pull/1422); this is the series the chart will draw; the chart itself follows.

## What changed

`ItemPracticeSummary.tempo_history` is replaced by `tempo_trend`.

`tempo_history` had no reader outside the preview fixtures, and it could not answer the one question a chart with gaps needs: **which sessions measured no tempo**. It only held the measured ones, so an unmeasured session was indistinguishable from no session at all, and a line drawn through it would silently interpolate across playing nobody measured.

`tempo_trend` holds one point per practised session, oldest first, with `tempo: None` where nothing was measured. The gap now has a position, so the chart can break the line there instead of joining two points across it.

Two rulings live in the core rather than the shell:

- a session with no measured tempo **keeps its slot** in the series;
- `has_trend` is **two or more** measured tempos, because below that there is no direction to read and the screen should say so plainly rather than draw a one-point chart.

Both are domain judgements about what the app is willing to claim, so neither belongs in Swift. What the shell derives from the series (the y-range, the "88 → 116" chip, the "7 of 9 sessions measured" line) is reading the data, not ruling on it.

Every number in the series was measured by construction: [#1422](https://github.com/jonyardley/intrada/pull/1422) keeps unevidenced defaults out of `achieved_tempo` in the first place, so nothing needs filtering here.

## Also here

Both from CLAUDE.md → Testing, "test fixtures for a type with many fields live in one place":

- `make_session_at`, which `make_session` now delegates to. The dated variant was being hand-rolled a third time for these tests.
- `ItemPracticeSummary.fixture` in `PreviewSupport`, which the six preview call sites now use. Adding this PR's field would otherwise have cost six edits; the next one costs zero.

## Tests

Five on the projection, plus one on the wire:

- ordering is oldest first, one point per session
- an unmeasured session is a `None` slot between its neighbours' numbers
- `has_trend` is false at zero and one measured tempo, true at two
- a series carrying a `None` round-trips on the real bincode FFI bridge, which is the shape most at risk on the positional wire ([#846](https://github.com/jonyardley/intrada/issues/846))

Both rulings were mutation-tested: relaxing `has_trend` to `>= 1` fails the threshold test, and reversing the sort fails the ordering and gap tests.

`just check` green (884), `just ios-fmt-check` green, `just ios-test-full` green.

## Not in this PR

The declared-target reference line. Mocked as option B on the [design canvas](https://claude.ai/code/artifact/b0aff8b0-b9e7-45bb-a209-fe8bb866f6fd); it needs no core change either way, and it costs the trend two thirds of its height, so it is a call for Jon on the screens PR rather than something to bake into the series.

Coverage: the projection and both rulings are covered by the five tests above. The only uncovered lines in the diff are the `PreviewSupport` fixture defaults, which are `#if DEBUG` and excluded (`ios/` is an ignored path).

Tier 3 (FFI bridge contract change, domain-sensitivity override).

### DIFF
--- a/crates/intrada-core/src/app.rs
+++ b/crates/intrada-core/src/app.rs
@@ -838,14 +838,14 @@ pub(crate) fn build_practice_summaries(
 
             if let Some(score) = entry.score {
                 record.2.push(ScoreHistoryEntry {
-                    session_date: session.started_at.to_rfc3339(),
+                    session_date: session_date.clone(),
                     score,
                     session_id: session.id.clone(),
                 });
             }
 
             record.3.push(TempoTrendPoint {
-                session_date: session.started_at.to_rfc3339(),
+                session_date: session_date.clone(),
                 session_id: session.id.clone(),
                 tempo: entry.achieved_tempo,
             });
@@ -3091,11 +3091,10 @@ mod tests {
         score: Option<u8>,
         tempo: Option<u16>,
     ) -> PracticeSession {
-        let now = started_at;
         PracticeSession {
             id: id.to_string(),
-            started_at: now,
-            completed_at: now,
+            started_at,
+            completed_at: started_at,
             total_duration_secs: 300,
             completion_status: CompletionStatus::Completed,
             session_notes: None,
@@ -3147,11 +3146,15 @@ mod tests {
 
     // ── The tempo trend (#1420) ──
 
-    fn tempo_trend_for(measured: &[Option<u16>]) -> crate::model::TempoTrendView {
+    /// `measured[i]` is the tempo of the i-th session chronologically. The
+    /// sessions are handed over newest first, so a projection that skipped its
+    /// own sort would emit the series backwards.
+    fn summary_for_tempos(measured: &[Option<u16>]) -> ItemPracticeSummary {
         let base = chrono::Utc::now() - chrono::Duration::days(30);
         let sessions: Vec<PracticeSession> = measured
             .iter()
             .enumerate()
+            .rev()
             .map(|(i, tempo)| {
                 make_session_at(
                     &format!("s{i}"),
@@ -3165,7 +3168,10 @@ mod tests {
         build_practice_summaries(&sessions)
             .remove("item-1")
             .expect("summary for item-1")
-            .tempo_trend
+    }
+
+    fn tempo_trend_for(measured: &[Option<u16>]) -> crate::model::TempoTrendView {
+        summary_for_tempos(measured).tempo_trend
     }
 
     #[test]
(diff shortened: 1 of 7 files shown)


jonyardley:
## Self-review

Code-reviewer subagent, `general-purpose`, against `origin/main..HEAD`, briefed that comment-policy violations are Blockers.

### What it found, and what I did

**Two tests passed for the wrong reason. Both fixed, both now mutation-proven.**

This is the finding worth reading. The reviewer deleted each of the two derivations this PR actually changed and the full 695-test suite stayed green:

1. **The oldest-first sort.** `tempo_trend_for` built its sessions in ascending date order, so the series came out ascending whether or not the sort ran. My own mutation test had *reversed* the sort (which the test caught) rather than *removed* it (which it did not) — the weaker mutation, and I called it verified. The fixture now hands sessions over newest first, so putting them in order is the projection's work rather than the input's.
2. **`latest_tempo`'s gap-skipping.** This line changed meaning in this PR, from "first of a newest-first measured-only list" to "last `Some` in an oldest-first list with holes". Replacing it with the naive `points.last().and_then(...)` broke nothing, and that naive version is a user-visible regression: an item whose most recent session measured no tempo would report no tempo at all, dropping the number the user earned the week before. Now pinned by `latest_tempo_is_the_newest_measured_not_the_newest_session`.

Confirmed independently: deleting the sort now fails two tests, and the naive `latest_tempo` fails one.

**Per-entry, not per-session.** The reviewer is right that `build_practice_summaries` accumulates per setlist entry, so an item appearing twice in one session emits two points at the same date. I have left the behaviour and corrected the doc comment, which said "per session". Reasoning: it matches the existing `session_count`, which has counted entries since long before this PR, and the alternative invents a merge rule (prefer measured over gap) to fix a state the builder's duplicate guards make unproducible today. Folding the two would be a change to `session_count` too, and that is its own PR.

**Preview fixtures held a state the core cannot emit** — a non-nil `latestTempo` beside an empty series. Inherited, but it would have bitten next: the chart PR records snapshot references against these. The four fixtures with a tempo now carry matching trends, including a gap and a single-measurement case, so the chart PR gets both interesting states for free.

**Three stale doc lines**, all inherited and all made contradictory by this PR: `status.md`'s Next entry described this PR's own work as unstarted, `roadmap.md` named the type this PR removes, and `comparison.md` called the chart the second PR when it is the third. Swept.

Also applied: dropped a pointless `let now = started_at` rebinding, and reused the `session_date` already in scope instead of a third RFC3339 allocation per entry (on the `score_history` push too, so the pair matches).

### Where I pushed back

**Not adding `measured_count` to the ViewModel.** The reviewer suggested exposing it since the core already computes it for `has_trend`, so the screen need not recount. I have left it out: it is derivable from `points` in one line, and a stored count is a second source of truth for a number that can drift from the series it describes. The three states the copy needs (never practised, practised but never measured, measured once) are all distinguishable from `points` and `has_trend` without it. `has_trend` earns its place because it is a *judgement* about what the app will claim; a count is not.

### Comment policy

Every added `//` and `///` is a section header, a contract on a bridge-crossing type, or a WHY with a citation. Two type docs were saying the same thing twice, so the pair is consolidated: the "gaps, not zeros, not interpolation" contract now lives once, on `TempoTrendView`.

### Gates

`just check` green (884 including the API crate), `just ios-fmt-check` green, `just ios-test-full` green.

Deferred items tracked: #1424.

PR #1425, opened 2026-08-28
The tempo trend (#1420), screens: draw the measured tempos and show where none were

jonyardley:
Third and last PR on [#1420](https://github.com/jonyardley/intrada/issues/1420). Stacked on [#1423](https://github.com/jonyardley/intrada/pull/1423) (the series); GitHub retargets this to main when that merges. Design: [Tempo Trend canvas](https://claude.ai/code/artifact/b0aff8b0-b9e7-45bb-a209-fe8bb866f6fd). Rulings: design-principles **T17**.

## The chart

`TempoTrend`, on the piece and exercise detail screens, above Recent sessions.

A line across the sessions that measured a tempo, on a **date axis** rather than session order, so a fortnight off reads as a fortnight rather than as one step.

**Where a session measured nothing, the line stops and restarts.** Nothing is joined across it, and nothing is drawn at zero. The unmeasured session takes a small tick **below** the axis, outside the plot: a mark on the baseline would read as a tempo of nothing, which is exactly the reading T16 exists to prevent. `7 of 9 sessions measured` sits under the plot, and is what keeps the breaks from looking like a fault rather than needing a paragraph to explain them.

**Under two measurements there is no plot.** One measurement reads `Measured once, at ♩ = 66 · no trend yet`; nothing measured at all draws no card, because nothing was asked of the user so there is nothing to report. Both follow the core's `has_trend` — the shell never rules on it.

**The `♩ = 96 → 104` chip is neutral**, never the teal `RecentSessions` uses on the way up. A faster tempo is not a better one: slowing down deliberately to fix a passage is practice, not a decline. This is the one place the two cards deliberately differ, and T17 says why.

## The call I made, in case you want it the other way

**The item's declared target is not drawn on the plot.** On the detail screen the Tempo row sits directly above the card, so `♩ = 108` is already on screen two lines up, and a reference line repeats it while costing the trend most of its height — an item at ♩ = 66 working towards 132 spends two thirds of the plot on empty space. Mocked as **option B** on the canvas if you want to see it; it needs no core change either way, so reopening the call is cheap.

## Accessibility

The header stacks instead of sharing a row at accessibility text sizes, where it was breaking `MEASURED` mid-word. The footer keeps the measured count and drops the end dates, because three labels cannot share that row and the count is the one that earns its place. Snapshotted at AX5 as well as default, since both can regress independently.

The whole card is one VoiceOver element: `Measured tempo, 96 to 104 beats per minute across 2 of 2 sessions`, spelling out the bpm rather than reading the ♩ glyph, matching `TempoFormatting.spoken`.

## Also here

- **The demo dataset records measured tempos.** It recorded none anywhere, so the seeded scheme could not show this feature at all. Hanon now carries ♩ = 96 → 104 across its two sessions and Clair de Lune one measurement, which gives the seeded app both states.
- **`ios-snapshots-record` now re-finds what it wrote** instead of reusing the list it deleted. On a *first* record that list is empty, so the reference it just wrote skipped optimisation and then failed the hygiene check — which is what happened here, twice, before I read the recipe.

## Verification

Driven on the simulator, not only in tests: both states confirmed on the seeded app (Hanon No. 1 with the two-point trend, Clair de Lune with the one-measurement line). Snapshots: two new component references (default + AX5) and one re-recorded screen reference, `testExerciseDetailLinkedFrom`, which gains the card because its fixture now carries a trend.

`just check` green, `just ios-fmt-check` green, `just ios-test-full` green, `just ios-snapshots-check` green.

Coverage: the drawing itself is `ios/`, an ignored path, so the patch report will cover only the demo-data change in `app.rs`, which the existing sample-data tests exercise. Expect no meaningful patch-coverage figure on this PR.

Tier 2 (new component following existing patterns; no bridge, DB, auth or migration surface — the bridge change was #1423).


jonyardley:
## Self-review

Code-reviewer subagent, `general-purpose`, against `claude/tempo-trend-chart-2915c6..HEAD`, briefed that comment-policy violations are Blockers.

### The finding worth reading

**The gap logic had nothing but a pixel watching it.** `runs` is the invariant this feature is named after, and its `count > 1` filter is exactly the kind of line a future tidy-up deletes without changing the snapshot much. The geometry and series arithmetic now live in `TempoTrendSeries`, tested directly across nine shapes including the degenerate ones (empty, one mark, all on one date, all one tempo): no run of line ever spans an unmeasured session, every measured session gets a dot and no unmeasured one does, and every point lands inside the plot. Confirmed both mutations fail: letting runs cross a gap, and dropping the scale floor.

**The scale floor is new, and it is an honesty fix.** The plot scaled to min...max of the measurements, so `♩ = 96 → 98` drew the identical dramatic climb as `♩ = 80 → 120`. That is the borrowed authority T16 exists to avoid, arriving through the back door. The scale now never spans less than 8 BPM, so small changes flatten. Nobody had ruled on this; it is now a T17 bullet.

### Also fixed

**VoiceOver was hearing nonsense on the no-trend path.** `Measured once, at ♩ = 108 · no trend yet` passed the ♩ glyph and the middle dot straight to the label, and VoiceOver reads neither, so it came out as "Measured once, at equals 108 no trend yet". The repo already knows this in two places (`ClickControl`, `TempoFormatting.spoken`) and the trend branch got it right; the no-trend branch did not. The only defect here a real user would have hit.

**The eyebrow was a hand-rolled clone of `Eyebrow`, and the third in the tree.** This card and `RecentSessions` now both use the primitive: they sit one directly above the other on this screen, so their disagreeing tracking (1.2 against `Eyebrow`'s 1.5) would have been visible side by side. `ReflectionSheet`'s hand-roll is a different surface with no adjacency, so it is #1426 rather than scope creep here.

**The card moved directly under the Key/Tempo rows.** T17 argues against drawing a target reference line on the grounds that the declared tempo is already on screen just above; the reviewer pointed out that was only true for exercises, because a piece has notes, tags, the chord chart and linked exercises in between. Moving the card makes the ruling true rather than aspirational, and puts the measured tempos where a musician would compare them to the target. Eight snapshot references re-recorded for the move and the tracking change.

**The preview fixtures now build through `tempoTrendDisplay`**, so the two new snapshots exercise the shipped mapping instead of a parallel copy. Worth noting the parallel copy was already wrong: its hard-coded six-date list silently truncated any fixture longer than six points, and the nine-point one was.

**Two T17 claims were false and are corrected.** The no-card ruling is the shell's, not `has_trend`'s, and it cannot be `has_trend`'s: `measured >= 2` cannot distinguish nothing-measured from measured-once. T17 now says which ruling lives where and what would move it to the core.

Plus: `TempoTrendMark` dropped its unread `id` and `Identifiable`; the seeded demo now shows a gap (Hanon's middle session measures nothing); six comments cut for restating the code or T17; two in-sentence em dashes removed from T17.

### Where I pushed back

**Kept the plot at 84pt at accessibility text sizes** rather than scaling it with the text. A chart is not text, growing it would push everything else off screen, and the numbers are in the chip and the count, both of which do scale. Noted rather than changed.

**Left `testExerciseDetailLinkedFrom`'s reference cut off mid-plot.** Its named coverage is the linked-from row at the top of the screen; the card has its own references. Framing that snapshot around a card it does not exist to test would be the tail wagging the dog.

### Gates

`just check` green, `just ios-fmt-check` green, `just ios-test-full` green, `just ios-snapshots-check` green. Driven on the simulator again after the move.

Deferred items tracked: #1426.

PR #1430, opened 2026-08-28
ios(picker): drop the chip tray from the related-exercise picker

jonyardley:
Jon, on the piece page: the chips under the "N related" header in the add-related-exercise sheet are redundant.

The tray repeated a selection the list rows already carry as a tick, on the same screen, one scroll apart. Chips are for input, not for mirroring a list that is already visible. The "N related" count line stays, and the empty hint with it.

## What changed

- `LinkedExercisePickerSheet`: `selectedTray` becomes `selectedCount` — the count line and the empty hint, no chips.
- `testLinkedExercisePickerSelectedTray` covered nothing but the chips, so it and its reference are deleted (no orphans).
- `testLinkedExercisePicker` re-recorded: same screen, minus the chip row.

Net -35 lines.

## Known trade-off

The chips were the only way to un-tick a selected exercise that search or a tag filter had scrolled out of view. The count still says it is there, and clearing the filter brings the row back. Judged not worth a second copy of the list.

## Verification

- `just check` green (885 tests).
- `just ios-fmt-check` green.
- `just ios-test` green; snapshot reference read back to confirm the chips are gone and the row ticks still carry the state.

## Coverage

Tier 1, snapshot-only surface. Net test deletion by design — the deleted test asserted a component that no longer exists. `ios/` is excluded from Codecov.

## Tier

Tier 1 (single component, no bridge / DB / auth / migration surface), so the review step is the gates plus this description per CLAUDE.md.

Deferred items tracked: none — all flagged items addressed inline.


jonyardley:
Two CI fixes and a doc update since the first push.

- **Snapshot Hygiene** was red: the record run aborted on a simulator launch flake before its `oxipng` step, so the reference went in at 242 KB against a 200 KB ceiling. Optimised, now 76 KB.
- **Comment density** was red at 0.20. The diff deletes comment lines net, but two surviving comment lines were *edited* (the doc sentence describing the deleted tray, and the section divider) against only ten added code lines, and the ratio has no signal at that size. Dropped the divider, which two members named `selectedItems` and `selectedCount` did not need. Now 0.10.
- `docs/status.md` records the second round of real-use findings and Jon's calls on them.

Worth knowing: the script's own failure message offers `SKIP_COMMENT_CHECK=1` as the bypass for justified cases, and the CI job does not honour it, so a branch pushed under the documented bypass can never go green. Filed as #1433.

Deferred items tracked: #1431, #1432, #1433.

Issue #1438, opened 2026-08-29
Spike: can we recover bar structure from a photographed lead sheet?


jonyardley:
Timeboxed spike named in [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md). **Not a phase, and deliberately not scheduled.**

## The question

`parse_chart` needs bar boundaries above all else. On a Real Book style lead sheet, OCR finds `Cmaj7` because chord symbols are text, and tells us nothing about barlines, repeats, endings or sections, because those are graphics. A bag of chord symbols with coordinates is not a chart.

Martinez-Sevilla et al. (2025), [Optical Music Recognition of Jazz Lead Sheets](https://arxiv.org/abs/2509.05329), state that chord notation is a score component not handled by existing OMR systems, against a corpus of 293 handwritten sheets. There is nothing to drop in.

Barlines are long vertical strokes, so recovering them geometrically is possible in principle. Whether it is reliable enough on a phone photo of a real page is the open question.

## Deliverable

A findings doc saying whether this is worth a phase at all, kept in `docs/research/`. The answer may well be no, and that is a useful result. Per the nothing-unread-stays-in-the-tree rule, the spike code is deleted and the findings kept.

## Do not start before

Phase D (#1387 part 2) has taught us how people actually photograph their charts. #1355 ships a photo store, which means the evidence arrives on its own.

Issue #1437, opened 2026-08-29
On-device model suggestions for a photographed page (Foundation Models)


jonyardley:
Phase C of [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md). Follows #1436, which does the same job with Vision OCR and geometry heuristics on every supported device.

## Outcome

On devices that have it, the on-device model fills the same confirm sheet more accurately than the heuristics can. No new screen, no new surface, no new user-facing concept.

## Scope

- Foundation Models `@Generable` guided generation into `SuggestedFields`, behind `if #available(iOS 26)` **and** `SystemLanguageModel.default.availability`. Populates the `suggested` field on `PageReading`; leaves it `None` everywhere else.
- **The clamp is the point.** Every field the model suggests must appear as a substring of the recognised OCR lines, or the core discards it and falls back to its own heuristic. A ~3B model asked to extract will sometimes produce a plausible composer that is not on the page. The clamp makes that structurally impossible, and it is a pure function tested in Rust with no device involved.
- No new `Effect`: this fills a field on the existing `RecognitionOutput` from #1436.

## Constraints

- The app targets iOS 17.0; Foundation Models needs iOS 26 and Apple Intelligence hardware (iPhone 15 Pro and later), a minority of the installed base. **This phase adds accuracy, never capability.** If it were removed entirely the feature must still work.
- On-device only. Not a cloud model, not as a fallback, not as an opt-in.

## Tier

Tier 2 if #1436 has already landed the effect and the clamp has a home; the domain-sensitivity override applies if the contract needs changing.

jonyardley:
Claiming this. Branch: `claude/piece-photo-phase-c-bf4618`. Also closes #1454 (the confidence question is this phase's call to make).

Issue #1436, opened 2026-08-29
Read a photographed page into title, composer and tempo (on-device OCR)


jonyardley:
Phase B of [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md). Follows #1355, which stores the photo and does no recognition at all.

## Outcome

Photograph the page, and the create form arrives pre-filled with the title, the composer and the tempo it could read, for the user to check and correct before saving. Nothing is written without that confirmation.

## Scope

- **New `Effect` family**, per the spec contract: `RecognitionOperation::ReadPage { photo_id }` resolving `RecognitionOutput`. The shell runs Vision text recognition and returns recognised lines with normalised geometry and confidence; it returns no domain types and makes no decisions. `Unsupported` is a normal outcome, not an error.
- **`read_fields(&PageReading) -> PhotoDraft`** in the core: a pure function, geometry heuristics only (largest text near the top is the title, a `by` / `Music by` line is the composer, a marking word or a bpm number is the tempo). This is where the tests live, and they need no device.
- **The confirm surface.** Open question 2 in the spec: its own review sheet, or pre-filling `ItemFormScaffold` with the values marked as suggestions. Design conversation before the screens PR. The second folds neatly into #1390.
- Low confidence is shown, not hidden.

## Constraints

- Fully useful on iOS 17 with Vision alone. #1387 adds model suggestions on top and must never become the mechanism.
- On-device only; the photo and everything read from it stay on the phone.
- Nothing recognised is ever saved silently, same rule as the scaffold preview.

## Tier

Tier 3: a new `Effect` crosses the FFI bridge. Two PRs, core first (the effect, `read_fields` and its tests), screens second.

jonyardley:
Picking this up. Branch: `claude/piece-from-photo-phase-b-669ce1`. Core PR first (the `RecognitionOperation` effect, `read_fields` and its heuristics), screens second after the open-question-2 design conversation.

jonyardley:
## Spec open question 2, answered

Jon, 2026-08-29: **pre-fill the existing add form with the recognised values, and let the user edit before saving.** Not a confirm sheet of its own.

[...]

Issue #1387, opened 2026-08-14
Chord chart entry: friendlier manual input, and a chart from a photo (on-device AI)


jonyardley:
Spun out of real use of the shipped chord-chart flow (step 2 of the adopted Stage 4 order, `docs/research/comparison.md`). The derivation is good; the *entry* is the most syntax-heavy surface in the app. Two parts, different sizes.

## 1. Friendlier manual entry (near-term)

Today: a raw text field with a format hint (`[Section]` labels, bars between `|` pipes, chords as written). Works, but it asks the user to learn a syntax. Explore, design-first (Claude Design before code):

- Structured entry: per-bar cells, a chord picker (root + quality) or chord keyboard, section chips.
- Live parse feedback per bar instead of reject-on-save.
- Keep the text path as the power-user fast lane; the parser already gives per-token errors.
- Input: friction notes filed from step 2's real use.

## 2. A chart from a photo, on-device (bigger bet)

Photograph a lead sheet or chart; the app reads the changes into a `ChordChart` for review before saving. Graduates "photo-of-your-own-chart OMR" from #1098 theme 4 into a concrete item.

- **Feasibility spike first.** Printed chord symbols are text: Vision framework OCR + the existing parser may get a long way. Handwritten or dense charts likely need an on-device model (Apple Foundation Models guided generation into the chart structure).
- **Constraints:** on-device only, the photo never leaves the phone (offline-first); review-before-commit, same no-silent-generation rule as the scaffold preview; read the *changes only*, never melodies (the iReal Pro copyright line, `specs/chart-to-scaffold.md`).
- The photo attachment groundwork in #1355 (photo aide-memoire on items) is adjacent: same capture surface, different consumer.

## Sequencing

Part 1 follows the friction findings from real use; part 2 sits with the big bets in the adopted order, after the quick wins. Design conversation first for both; tier per the surfaces touched when picked up.

jonyardley:
Designed: [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md) covers this alongside its siblings.

- **#1355** is phase A: the photo, the file store, `Item.photo_id`, migration v16, capture and display. No recognition of any kind, and useful on its own.
- **#1436** is phase B: Vision OCR into title, composer and tempo, via a new `RecognitionOperation` effect so the core still owns interpretation.
[...]

PR #1439, opened 2026-08-29
docs(specs): design for adding a piece from a photo

jonyardley:
Design doc for photographing a page and getting a piece out of it. Docs only: no Rust, Swift or config touched.

Covers #1355 (a photo on an item) and #1387 part 2 (a chart from a photo), and opens #1436, #1437 and #1438 for the parts neither issue covered.

## Deviation worth flagging

CLAUDE.md says a Tier 3 spec rides with the first implementation phase, not its own PR. This one does not, because Jon asked for a design document in the repository to work off *before* implementation starts, and a spec sitting on an unmerged branch is not in the repository in any useful sense. Phase A (#1355) should still open with a normal core-then-screens PR pair; the spec will already be on main.

## What it says

Four phases, each its own issue and PR pair, each useful alone.

- **A** (#1355): the photo, the file store, `Item.photo_id`, migration v16, capture and display. **No recognition of any kind.** A photo of the page you are practising from earns its place with nothing attached, and it also tells us what people actually photograph before phase D commits to a parser.
- **B** (#1436): Vision OCR into title, composer and tempo.
- **C** (#1437): on-device model suggestions, where the hardware allows.
- **D** (#1387 part 2): a chart from a photographed *text* chart, into the existing `ChordChartEditSheet`.

Lead sheets over a stave are **not** a phase. Chord symbols are text so OCR reads them, but barlines are graphics, so you get a bag of chord symbols with coordinates and no bar structure, and `parse_chart` needs bars above all else. [Martinez-Sevilla et al. (2025)](https://arxiv.org/abs/2509.05329) say chord notation is not handled by existing OMR systems, against a corpus of 293 handwritten sheets. That half is a timeboxed spike (#1438) whose deliverable is a findings doc, possibly saying no.

## The two decisions worth reviewing

1. **Recognition is an `Effect`, not Swift logic.** The shell owns VisionKit, Vision and Foundation Models and returns recognised lines with normalised geometry and confidence. The core decides what any of it means, in one pure `read_fields` function. This is what keeps the dumb-pipe rule intact with an on-device model in the loop.
2. **The model may choose, never invent.** Every field the on-device LLM suggests must appear as a substring of the OCR lines, or the core discards it and falls back to its own heuristic. A ~3B model asked to extract will sometimes produce a plausible composer that is not on the page. The clamp makes that structurally impossible, and it is testable in Rust with no device.

Two further constraints the spec pins: the app targets iOS 17.0 while Foundation Models needs iOS 26 and Apple Intelligence hardware, so heuristics are the mechanism and the model is only ever an enhancement; and the photo is a file on disk with only its id on the item, never a blob in GRDB, because that table is loaded whole and the device is the only copy of the data.

## Verified against the code, not assumed

- `parse_chart`s bar-and-pipe grammar and its need for bar boundaries (`chart.rs:147`).
- `SetlistEntry` denormalises item fields rather than embedding `Item` (`session.rs:56`), so a new `Item` field does **not** invalidate the `ActiveSession` crash-recovery blob. The #1223 class of trap does not apply here.
- Current GRDB head is v15, so the migration is v16.
- There is no Vision, Photos, camera or image handling anywhere in `ios/` today.

## Gates

`just check` green: 891 tests, typos and cargo-shear clean. No `ios/` files touched, so `just ios-fmt-check` and `just ios-test-full` do not apply.

Coverage: not applicable, docs only.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -203,3 +203,13 @@ audit backlog and its five-phase build order.
   (#1103) moved to now.
 - Follow-up: port the wire-pin test technique (per-variant bincode
   fingerprint) to the `ActiveSession` crash-recovery blob (#1345).
+- **Adding a piece from a photo is designed, not built**
+  ([`specs/piece-from-photo.md`](../specs/piece-from-photo.md)). Four phases:
+  #1355 stores the photo and does no recognition; #1436 reads title, composer
+  and tempo with Vision OCR through a new `RecognitionOperation` effect; #1437
+  adds on-device model suggestions, clamped so the model may only choose text
+  that appears in the OCR; #1387 part 2 reads a photographed *text* chart.
+  Phase A is useful alone and is the entry point. The load-bearing finding:
+  chord symbols on a stave are text so OCR reads them, but barlines are
+  graphics, so a lead sheet gives no bar structure and `parse_chart` needs bars
+  above all else. That half is a spike (#1438), explicitly not a phase.
--- a/specs/README.md
+++ b/specs/README.md
@@ -25,6 +25,12 @@ Companion documents outside this folder:
 [`docs/roadmap.md`](../docs/roadmap.md) (direction and phase order) and
 [`docs/status.md`](../docs/status.md) (what's in flight).
 
+## Planned — designed, not built
+
+| Spec | Scope |
+|---|---|
+| [`piece-from-photo.md`](piece-from-photo.md) | Adding a piece from a photograph of the page: storage, then OCR, then on-device extraction (#1355, #1387) |
+
 ## Shipped record — verify against the code before extending
 
 Builder-era specs whose surfaces returned with the restored session builder.
--- a/specs/piece-from-photo.md
+++ b/specs/piece-from-photo.md
@@ -1,0 +1,262 @@
+# Add a piece from a photo
+
+> Tier 3 by domain sensitivity: a new `Item` field crosses the FFI bridge, a
+> new GRDB migration lands on the only copy of the user's data, and a new
+> `Effect` family is added. Issues [#1355] (photo on an item) and [#1387]
+> part 2 (a chart from a photo); [#1390] (one-pass create) is adjacent and
+> should land after, or reuse, phase A's capture surface. Roadmap: Layer 1
+> (Capture), `pillar:plan`. Graduates "photo-of-your-own-chart OMR" from
+> [#1098] theme 4.
+>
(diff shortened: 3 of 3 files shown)


jonyardley:
## Self-review

**The code-reviewer subagent was not run.** This session carries an explicit instruction not to call the Agent tool unless asked, so `requesting-code-review` was skipped and I reviewed the document myself. Worth a second pair of eyes on the contract before phase A opens, given it is a Tier 3 bridge design.

### Applied inline (would have been blockers)

1. **`PhotoDraft` was referenced by the contract and never declared.** The one function the whole design hangs on returned an undefined type. Now declared, and carrying per-field provenance (`DraftSource::Recognised` vs `Suggested` plus confidence) so the confirm sheet can show a shaky read differently from a clean one, which decision 7 assumed and the contract did not support.
2. **No round-trip precondition.** Every type here crosses the bincode bridge and the spec said nothing about `assert_round_trips`. CLAUDE.md makes that a build precondition and #846 is the reason. Now stated, along with `photo_id` appending after `variants` because the wire is positional.
3. **The offline-first checklist was implied, not answered.** Now walked invariant by invariant, so the phase A PR has something to check itself against rather than re-deriving it.

### Applied inline (important, not blocking)

4. **Nobody reaps orphan photo files.** Decision 2 leaves bytes behind on a tombstone, which is right for the user and unbounded for the disk. Raised as open question 5, to be answered before phase A ships rather than discovered later.
5. **HEIC vs JPEG was asserted in passing** in decision 1 with no reasoning. Now open question 6: cheap to decide now, expensive once photos exist on real devices.
6. **The day estimates were stated too confidently.** No slice in this codebase has used Vision or a camera before, so they have no comparable to rest on. Now flagged as rough, with only phase A worth planning against.

### Standing concerns, not fixed

- **Phase D is the least certain thing here and the spec says so, but it is also the phase most likely to be what Jon actually wants.** The honest sequencing is that phase A produces the evidence (what people photograph) that tells us whether D is worth building, which is why A is deliberately useful with no recognition attached. If real photos turn out to be lead sheets rather than text charts, D does not happen and #1438 is the only route.
- **Open question 2 (confirm sheet vs pre-filled form) is load-bearing for phase B and unanswered.** It needs the design conversation before B's screens PR, not during it.

### Gates

`just check` green: 891 tests, typos and cargo-shear clean. No files under `ios/` touched, so the native-iOS tier does not apply. Docs only, so no coverage line.

Deferred items tracked: #1436, #1437, #1438

Issue #1440, opened 2026-08-29
Pickers disagree on how they filter: shared query vs local state


jonyardley:
The app now has two opposite answers to "how does a picker filter its list?".

- `AddToSessionSheet` and `AddRelatedExerciseSheet` reuse `BrowseControlsBar`, which writes the core's shared `ListQuery` (the same one the Library screen uses). The related-exercise sheet scopes that query on appear and restores it on dismiss (#1103) so the scoping doesn't follow the musician back to the Library.
- `LinkedExercisePickerSheet` keeps its filter state shell-local and does not touch the shared query, and its doc comment says so deliberately.

Both work, but a musician gets different behaviour depending on which picker they opened, and the next picker someone builds has no obvious precedent to copy.

Pick one:

- teach `BrowseControlsBar` to drive a caller-supplied binding, so a picker can filter without writing the shared query, and move the two add sheets onto it; or
- accept the shared-query route everywhere and retire the local one.

Raised in the self-review of the #1103 PR.

jonyardley:
Claimed. Branch: `claude/picker-filter-consistency-1440`.

PR #1441, opened 2026-08-29
ios(picker): keep the related-exercise sheet open, with search (#1103)

### LINKED ISSUE #1103: AddRelatedExerciseSheet: align with the add-sheet selection model (search + persistent multi-select)
From the 2026-07-15 builder UX audit (out of scope for the direct-manipulation PR):

`AddRelatedExerciseSheet` (opened from a block's "Add a related exercise" row) is one-shot tap-and-dismiss with no selected state and no search/sort/filter, while `AddToSessionSheet` one tap away is persistent toggle multi-select with the full `BrowseControlsBar`. Adding three exercises to a block means opening the sheet three times, and a long exercise library is unsearchable in it.

Align the two: persistent multi-select with added-state feedback (checkmark + stroke), keep the sheet open, and give it at least search. Candidates stay filtered to exercises not already in the setlist.

jonyardley:
Closes #1103.

## What changed

Opening a block's **Add a related exercise** row used to give a one-shot picker: tap an exercise, the sheet closed. Adding three meant opening it three times, and a long exercise list had no search in it, while **Add to session** one tap away was persistent multi-select with the full browse bar.

The two now behave the same way:

- The sheet stays open. Each row shows whether it is in the block (checkmark and accent stroke), and tapping an added row takes it back out.
- It carries the shared browse bar: search, sort and tag filter. No piece/exercise menu, since the sheet is exercises only and that menu could only empty its own list.
- Candidates are the exercises already in this block, plus any not in the setlist at all.
- The shared library query is scoped to exercises while the sheet is open and restored on dismiss, so a filter the Library left behind cannot narrow the sheet invisibly, and the scoping does not follow the musician back to the Library.

Shell-only: `SessionEvent::AddExerciseToBlock` is already idempotent on block membership, so no Rust changed and the bridge is untouched.

The added-state row treatment moved into a shared `SelectableLibraryRow`, used by both sheets, and both dropped their hand-rolled errorSeq/haptic pair for the existing `store.send(_:onSuccess:)` helper.

## Verification

- Driven on the simulator end to end: added an exercise (sheet stayed open, row flipped to added), removed one, searched, opened the tag filter from inside the sheet and confirmed the exercise scope survived the nested sheet, and confirmed the Library filter came back as it was on dismiss.
- New snapshot `testAddRelatedExerciseSheetAdded` covers the added state (one row in the block, one not); the existing sheet reference was re-recorded.

Coverage: no Rust changed, so no patch coverage is expected. iOS is in Codecov's ignore list; the gate for this change is the snapshot pair plus the simulator run above.


jonyardley:

**Verdict: ready with fixes** — both fixes applied inline, second commit on the branch.

**Strengths it called out**
- Shell-only, no bridge or persistence surface; candidate set correct including ungrouped entries (they count as "in the setlist elsewhere", so they stay hidden), and pieces can never appear, so the block's own piece can't be removed here.
- Extraction over cloning: `SelectableLibraryRow` collapses the duplicated overlay, `BrowseControlsBar` gained a parameter rather than a parallel bar.
- The new snapshot genuinely constrains the added state (one row in the block, one not) rather than re-photographing the empty case; both references optimised and mapped to a test.

**Fixed inline**
1. *Important* — the sheet inherited the Library's search text while `BrowseControlsBar` starts with its search bar hidden, so a word the musician never typed here could narrow the list with nothing on screen explaining it. The sheet now opens with no text and no tags, only the exercise scope.
2. *Important* — both sheets hand-rolled the errorSeq/haptic pair (and duplicated its #846 comment) instead of `store.send(_:onSuccess:)`, which already carries that rule. Both moved onto the helper.
3. *Minor* — empty state ignored tags; it now says "No exercises carry those tags." with the filter glyph filled. Verified on the simulator.
4. *Minor* — the four-line comment above the query scoping is trimmed to the constraint.
5. *Minor* — checked the nested-sheet worry on the simulator: opening the tag filter from inside the sheet does not fire `onDisappear`, and the exercise scope survives it.

**Not taken**
- Renaming the sheet title to "Related exercises". It matches the row that opens it, which is the more useful anchor.

Deferred items tracked: #1440

Issue #1442, opened 2026-08-29
Reap photo files nothing references any more


jonyardley:
Follow-up to #1355 phase A (spec: `specs/piece-from-photo.md`, key decision 2).

Phase A never deletes a photo file. Replacing a photo, clearing it, or deleting
the item all leave the image on disk. That is deliberate: an orphan file costs
disk, an eagerly deleted one costs the user their photo, and on the free offline
tier the device is the only copy.

The concrete hazard that settled it: the core would have emitted `SaveItem` and
`DeletePhoto` from one `Command::all`. The shell processes persistence ops in
order and unconditionally, so a `SaveItem` that throws (disk full, DB locked)
still lets the delete run. The core then reloads from the store, reads back the
row as it still is on disk, and the model points at a file that no longer
exists. The user's photo is gone and the reference is permanently broken.

So files accumulate. This issue is the pass that clears them.

**The shape**: after hydration the core emits `PrunePhotos { keep: [photo_id] }`
and the shell deletes everything in `Application Support/photos/` that is not on
the list.

**What has to be true first**: pruning on that signal is only safe if hydration
is provably complete. If the load path can ever resolve with a partial item
list, the sweep deletes photos of items that had not arrived yet. Audit the
hydration path for partial-result cases before wiring this up.

**Also decide here**: whether tombstoned items' photos belong in `keep`. There is
no undelete surface today, so they probably do not, but that is a data-loss
decision and deserves to be made explicitly rather than fall out of the query.

Issue #1355, opened 2026-08-14
Create item: attach a photo as an aide-memoire (score chart, lesson notes)


jonyardley:
From Jon's v0.7.0 audit. Iteration 1: exactly one photo per item, captured or picked at create/edit, shown on the item detail. Future iterations (out of scope, keep in mind for the data shape): multiple photos, photo management, versions.

Constraints: offline-first (photo stored on device; sync deferred), photo row/column carries updated_at + deleted_at, client-minted id. Touches FFI + GRDB schema, so Tier 3: short spec first, and Claude Design for the capture/display UX.

jonyardley:
Designed: [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md) covers this alongside its siblings.

- **#1355** is phase A: the photo, the file store, `Item.photo_id`, migration v16, capture and display. No recognition of any kind, and useful on its own.
- **#1436** is phase B: Vision OCR into title, composer and tempo, via a new `RecognitionOperation` effect so the core still owns interpretation.
- **#1437** is phase C: on-device model suggestions, clamped so the model may only choose text that appears in the OCR, never invent it.
- **#1387** part 2 is phase D: a chart from a photographed *text* chart, into the existing `ChordChartEditSheet`.
- **#1438** is the spike for lead sheets over a stave, which no shippable OMR handles today and which is explicitly not a phase.

The load-bearing finding: chord symbols on a stave are text, so OCR reads them, but barlines are graphics, so it cannot tell you which bar they are in. `parse_chart` needs bars above all else. A typed or handwritten *text* chart is a different and much more tractable problem.

jonyardley:
Claiming this. Branch: `claude/piece-photo-storage-phase-a-b05593` (Phase A: storage + core, no recognition). Spec lands with the Phase A core PR per the Tier 3 workflow.

jonyardley:
Screens on `claude/claim-1355-capture-display-61dd5c` — #1449. Core landed in #1443.

jonyardley:
Landed: core in #1443, screens in #1449.

A photo of the page you practise from now sits on the item detail, between the key/tempo rows and the chord chart. Scan the page (VisionKit's document scanner) or choose one from the library; tap it to read it full screen with pinch and double-tap zoom. Nothing deletes the file — replacing, clearing and deleting the item all leave the bytes on disk for the reaping pass, #1442.

[...]

Issue #1083, opened 2026-07-15
Exercise steps (variants): one ladder mechanism for keys, levels and inversions


jonyardley:
**Shipped apart from step-management polish** (state trued up 2026-08-14; this issue stays open for the polish only).

"Land on the 3rd, then add a note each round" is a ladder. So is "all twelve keys", and so is "shells in each inversion". One mechanism: an exercise owns an ordered list of steps (core name: variants), each scored independently; progress means advancing the step, not just polishing the rating. Multi-key practice shipped as this mechanism's first preset (closed #46).

Spec: `specs/exercise-variants.md`.

## State

- [x] **Core + data:** `Variant { id, label, position }` owned by exercise items; `variant` child table with per-row `updated_at`/`deleted_at` (sync-ready, invariant 2) + upgrade-path test; `variant_id` in the session-entry codec, serde-defaulted; real-bridge tests (#846 class). "Current step" derived: first step not yet solid. — #1112, #1118
- [x] **iOS:** exercise-detail Steps section (ladder with per-step rings); reflection-sheet step picker defaulting to the current step. — #1123
- [x] **Twelve-key preset + key grid** (closed #46). — #1121
- [ ] **Step management polish:** rename, reorder, archive.

## Key decisions (2026-07-15)

- Local-first-only until sync: the online path gets no step support yet; a test asserts online mode ignores steps gracefully, so invariant 6 is consciously scoped rather than silently broken. Server tables arrive with the sync engine.
- Naming: users see "Steps" (or "Keys" for the key preset); "variants" is the core's name, never on screen. UI state label is "Solid", not "consolidated".
- Step scope capped: label + position + score history. No per-step tempo, notes or targets in v1; every extra field is a migration on the tier where migrations are dangerous.

PR #1443, opened 2026-08-29
core(photo): store one photo per item, on the device (#1355)

jonyardley:
Phase A core of [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md) (#1439), for #1355: the aide-memoire photo an item carries. **No recognition of any kind** — that is phases B to D. Useful alone: a photo of the page you are practising from earns its place with nothing else attached.

## What landed

- `Item.photo_id: Option<String>`, appended after `variants` with `#[serde(default)]` because the bincode wire is positional.
- `ItemEvent::SetPhoto { id, photo_id }` / `ClearPhoto { id }`, local-first only. Kept out of `UpdateItem`: its three-state `Option<Option<T>>` is the fiddliest encoding on the bridge, and `AddTags`/`RemoveTags` set the precedent.
- `LibraryItemView.photo_id` — the projection the screens PR reads, so it will not have to reopen core (the #1256 phase C lesson).
- `validate_photo_id`: must parse as a ulid. Load-bearing rather than tidy — the id becomes a path component in the shell, so a non-ulid is a traversal out of the app container, and the core is the only layer that can refuse it before it is stored.
- Migration **v16_item_photo**, a nullable `photo_id TEXT`, plus the GRDB row↔Item codec.

## Nothing deletes a photo file

Key decision 2 in the spec, and worth stating plainly because the first version of this PR got it wrong. Replacing a photo, clearing it, and deleting the item all leave the image on disk.

The concrete hazard: the core had emitted `SaveItem` and `DeletePhoto` from one `Command::all`. The shell processes persistence ops in order and unconditionally, so a `SaveItem` that throws still lets the delete run; the core then reloads and reads back the pre-write row, and the item points at a file that no longer exists. On the free offline tier that file is the user's only copy. An orphan costs disk, an eager delete costs the photo. Reaping is #1442, wholesale.

`PersistenceOperation::DeletePhoto`, `PhotoStore` and `FilePhotoStore` went with it — nothing reads them until the screens PR writes files, and nothing unread stays in the tree.

## Offline-first checklist

- [x] **1** New writes go through the persistence `Effect`, not HTTP. Both events assert zero `Http`; both refuse the online path.
- [x] **2** `photo_id` sits on `item`, which carries `updated_at` + `deleted_at`; no hard delete of anything. At one photo per item the photo is an attribute (like `composer`), so item-granular LWW covers it; multiple photos is a `photo` child table migration, as `variant` was for step ladders (#1083).
- [x] **3** Client-minted ulid — the shell mints it (it needs one before it writes the file), the core validates it.
- [x] **4** All lifecycle decisions are in the core; Swift decides nothing.
- [x] **5** No new fake `Ack`. The one op that could have surfaced a spurious `Failed` (photo reclamation) no longer exists.
- [x] **6** New code, local-first only. `Add`/`Update` are untouched by photos, so their online branches cannot drop one.
- [x] **7** Migration appended, additive, nullable, with an upgrade-path test seeded at v15.

**Crash-recovery blob untouched**: `ActiveSession` → `Vec<SetlistEntry>`, which denormalises item fields rather than embedding `Item`, so the positional-bincode snapshots on devices stay valid and the key needs no bump. Verified by reading the graph (spec decision 3).

## Testing

Every assertion was mutation-tested by **deleting** the line it guards (or, where the test asserts an absence, by substituting the naive version a future reader would write). Eight mutations, eight kills:

| Mutation | Test that died |
|---|---|
| delete the same-id guard | `setting_the_same_photo_twice_writes_nothing_the_second_time` |
| delete `validate_photo_id` | `set_photo_rejects_an_id_that_is_not_a_ulid` |
| delete the `local_first` guard | `photo_events_are_refused_online` |
| re-add eager reclamation on item delete | `deleting_an_item_leaves_its_photo_file_alone` |
| delete `photo_id: item.photo_id.clone()` | `view_carries_the_photo_id_through_to_the_screens` |
| delete `photo_id = excluded.photo_id` | `testV16PhotoIdRoundTripsAndClears` |
| delete migration v16's `ALTER TABLE` | `testV16AddsPhotoIdColumnWithExistingItemsIntact` |
| delete `item.photo_id = next` | `testRealBridgePhotoIdCrossesTheWireBothWays` |

The last one is a **real-bridge** round-trip, not `assert_round_trips`: Swift serialize → core → projection → Swift deserialize, both set and clear. A stub bridge cannot catch a wire break (#846), and both the Rust-only and Swift-only mutations above kill it.

Gates: `just check` (903 tests), `just ios-fmt-check`, `just ios-test-full` — all green locally on the post-merge tree.

Coverage: expect full patch coverage on the new core paths — every branch of `set_photo` and `validate_photo_id` has a test, as does the view projection. `ios/` is Codecov-ignored, so the GRDB codec and migration are covered by the Swift tests instead. No expected gaps.

## Not in this PR

- Capture and display. Claude Design first, then the screens PR (`VNDocumentCameraViewController` plus a library picker, display on `LibraryDetailScreen`).
- Reaping orphaned files: #1442.

#1355 stays open until the screens land.


jonyardley:
## Self-review (code-reviewer subagent, per CLAUDE.md Always(5))

Reviewed `c766236..49f857a` against `specs/piece-from-photo.md`, the offline-first checklist and the comment policy. Findings and what I did.

### Blockers, all fixed

**B1 — this branch overwrote a spec that had merged onto main the same day.** The branch was cut from `cc4a27c`; #1439 landed `specs/piece-from-photo.md` at `ec1a00b` with a reviewed four-phase design, and this branch wrote a different 253-line document at the same path. My claim check ran `gh pr list --search "1355"` and `gh issue view 1355 --json closedByPullRequestsReferences`, both empty, and neither surfaces a *merged docs PR that references an issue without closing it*. Branching from `cc4a27c` rather than fresh `origin/main` is what let it stay invisible.

Fixed by merging main and taking its spec byte-for-byte (`b097dc0`), then realigning the code to it (`824b1e6`). Worth noting for the next session: the claim protocol as written does not catch this class, and a `git log origin/main -- specs/` before writing a spec would have.

**B2 — a failed `SaveItem` plus a successful `DeletePhoto` destroyed the user's photo.** The core emitted both from one `Command::all`; the shell runs persistence ops in order and unconditionally, so a throwing write still let the delete run, the core reloaded the pre-write row, and the item ended up pointing at a file that no longer existed. On the free tier that file is the only copy. Real, and the reviewer's reasoning was correct end to end.

Fixed by removing eager reclamation entirely, which is what the merged spec's key decision 2 already said. All reaping moves to #1442. `PersistenceOperation::DeletePhoto`, `PhotoStore` and `FilePhotoStore` went with it — nothing reads them until the screens PR writes files.

**B3 — comment-policy violations.** Four restatements deleted (`set_photo`'s doc, `LibraryItemView.photo_id`'s doc, the `ItemFormModel` note referencing this PR, `PhotoStore`'s 8-line header, which went with the file). Two over-cap comments trimmed to the WHY. The reviewer was right that the bridge rationale had been restated five times; it now lives in the spec with one-line pointers.

### Important, all fixed

**I1 — branch behind main.** Main added a `LibraryItemView(` site in `PreviewSupport.swift` with no `photoId:`. Confirmed: it broke the build after the merge, now fixed. Gates re-run green on the merged tree.

**I2 — the online path silently dropped a photo.** `CreateItem.photo_id` was stamped optimistically, posted as `None`, and clobbered by the server's echo. Fixed by removing `CreateItem.photo_id` — the merged spec scopes phase A to `SetPhoto`/`ClearPhoto` on an existing item, so the field was out of scope as well as buggy. With it gone, `photo_id` is unreachable in online mode, so the `ItemUpdated` wholesale-replace the reviewer also flagged has nothing to clobber.

**I3 — reclamation failure surfaced a user-facing storage error and refetched the library.** Gone with the reclamation itself.

**I4 — no real-bridge round-trip.** Correct and the most useful finding after B2: `assert_round_trips` is serde-in-Rust and cannot catch a Swift serializer mismatch, which is exactly #846. Added `testRealBridgePhotoIdCrossesTheWireBothWays`, covering Swift serialize → core → projection → Swift deserialize for both set and clear. Verified it bites from both sides: dropping the Rust projection kills it, and so does making the handler ignore the id.

### Nits

- `Store.swift` photo-store default inverting the in-memory convention: gone with `PhotoStore`.
- `FilePhotoStore` trusting the core's validation: gone; worth re-raising in the screens PR, which adds the write path.
- The same-id guard returning before `updated_at` moves: kept deliberately, and now has a test asserting `updated_at` does *not* move, so an unchanged item cannot win a later LWW merge. Photo ids are never reused (the shell mints a fresh ulid per capture) — that belongs in the screens PR's contract.
- `updated_at > before` timing dependency: accepted; the fixture is built microseconds earlier and the assertion is behavioural in intent.

### Pushback

None. Every finding held up on inspection, including the two I would have argued with on first read (B2's ordering claim — confirmed against `Store.process`, which is synchronous and in-order; and B1's severity — confirmed by diffing against `origin/main`).

### On the spec

The reviewer's judgement that the merged spec's decisions 1 and 4 are sound, and that item-delete reclamation was the wrong call, matches where the code ended up. No spec change proposed.

Deferred items tracked: #1442

PR #1444, opened 2026-08-29
ios(picker): search the same fields the Library searches (#1440)

### LINKED ISSUE #1440: Pickers disagree on how they filter: shared query vs local state

jonyardley:
Closes #1440.

## What changed

Two pickers disagreed about what search means. On a piece, **Choose one** opens **Add exercises**, whose search read only the exercise's title and its key/tempo line. The Library's search, and the two sheets in **Build session**, match title, composer, notes and tags. So typing a composer name or a tag in that one sheet found nothing, for no reason a musician could see.

- `LibraryItemView.matchesSearch` puts the core's four fields in one named, tested place, and the picker uses it. It no longer matches the key/tempo line, which the Library never did: typing `108` in that sheet used to find Hanon and now doesn't. That is the point of the change, but it is a loss worth naming.
- The two **Build session** sheets now put the Library's own filter back when they close, via a small `libraryQueryScope` modifier, so filtering inside a sheet no longer follows you out.

## What this deliberately does not do

The issue asked to pick one mechanism for all three pickers. I tried the shared-query route and abandoned it for a concrete reason: `LibraryScreen` resolves a pushed detail screen by looking the item up in the query-filtered list, so a sheet opened from a piece's detail that writes that query pops the screen underneath itself. Typing one character in the picker closed the sheet and the piece.

Jon and I weighed the remaining options against what a musician actually gets. Unifying the plumbing costs either a core change (a second query slot on the bridge) or a morning of shell churn across three working sheets, and changes nothing anyone can see. So this PR fixes the behaviour and leaves the mechanisms alone: the picker still filters locally, the two add sheets still browse the shared query. The screen-popping fragility is real but unreachable today, since nothing shipped writes the query from a detail screen; it is tracked separately.

## Verification

- Driven on the simulator: opened **Choose one** on Nocturne, searched `warm` and got Hanon by its tag, which found nothing before. Confirmed the sheet stays put while searching, and that the **Build session** sheets leave the Library's filter as they found it.
- New table test, mutation-tested by deleting each field branch of `matchesSearch` in turn and confirming a case fails each time. The first version of the test did not do this — its fixture had the same word in title and composer, so the title branch could be deleted with the suite still green. Caught in self-review.

Coverage: no Rust changed. iOS is in Codecov's ignore list; the gate here is the new table test plus the simulator run above.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -29,6 +29,16 @@ audit backlog and its five-phase build order.
 
 ## Recently landed
 
+- #1440 — **search means the same thing in the exercise picker as in the
+  Library**. Choosing an exercise for a piece (**Choose one** on the piece)
+  opens a sheet whose search only read the exercise's title and its key/tempo
+  line, so a composer name, a note or a tag found nothing there while the same
+  words worked everywhere else. It now matches the fields the Library matches.
+  The two add sheets in Build session also put the Library's own filter back
+  when they close, instead of leaving it as the sheet left it. The deeper
+  tidy-up — the two pickers are still built differently, one on the shared
+  query and one on its own state — was weighed and left, since it costs a
+  morning and a musician cannot see it.
 - #1103 — **the related-exercise picker now behaves like the add-to-session
   sheet**. Adding three exercises to a block used to mean opening the sheet
   three times, and a long exercise list had no search in it. The sheet now
--- a/docs/status.md
+++ b/docs/status.md
@@ -33,12 +33,13 @@ audit backlog and its five-phase build order.
   Library**. Choosing an exercise for a piece (**Choose one** on the piece)
   opens a sheet whose search only read the exercise's title and its key/tempo
   line, so a composer name, a note or a tag found nothing there while the same
-  words worked everywhere else. It now matches the fields the Library matches.
+  words worked everywhere else. It now matches the fields the Library matches,
+  and no longer matches the key/tempo line, which the Library never did.
   The two add sheets in Build session also put the Library's own filter back
   when they close, instead of leaving it as the sheet left it. The deeper
-  tidy-up — the two pickers are still built differently, one on the shared
-  query and one on its own state — was weighed and left, since it costs a
-  morning and a musician cannot see it.
+  tidy-up was weighed and left: the two pickers are still built differently,
+  one on the shared query and one on its own state, and unifying them costs a
+  morning for nothing a musician can see.
 - #1103 — **the related-exercise picker now behaves like the add-to-session
   sheet**. Adding three exercises to a block used to mean opening the sheet
   three times, and a long exercise list had no search in it. The sheet now
--- a/ios/Intrada/Core/LibraryItemView+Search.swift
+++ b/ios/Intrada/Core/LibraryItemView+Search.swift
@@ -1,10 +1,8 @@
 import SharedTypes
 
 extension LibraryItemView {
-  /// The shell's copy of the core's free-text match (`apply_query_filter` in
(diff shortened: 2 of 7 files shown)


jonyardley:

**Verdict: ready with fixes** — all applied, second commit on the branch.

**Blocker it found, and it was right**
- The headline test did not constrain the headline change. The fixture had "Hanon" in both the title and the composer, so every case still passed with the title branch deleted from `matchesSearch` — the one field the old predicate already had was the one field the test left free. Exactly the shape CLAUDE.md was amended twice to prevent. Fixed by making each field carry a word none of the others do, then deleting each branch in turn and confirming a case fails each time (title fails three).

**Also fixed**
- `libraryQueryScope` captured the query on every `onAppear`. A second appear before `onDisappear` would have stored the already-cleared query as the one to restore, quietly losing the Library's filter. It captures once now.
- Two em dashes mid-sentence in `docs/status.md` and the test's doc comment.
- Dropped a `///` on `libraryQueryScope()` that restated the type's own doc comment.
- Softened the claim in `matchesSearch`'s comment: it mirrors the core's *fields*, while the comparison is the shell's localised one, which is not bit-for-bit the core's `to_lowercase`.

**Checked and confirmed by the reviewer**
- `matchesSearch` agrees field-for-field with `apply_query_filter` (`app.rs:1136`).
- Dropping the query-scoping from `AddRelatedExerciseSheet` is safe: it filters `itemType == .exercise` itself and its type menu is hidden.
- Both scoped sheets are presented from `SessionBuilderScreen`, so the detail-screen popping hazard cannot fire there, and clearing widens rather than narrows in any case.
- `metaLine` still has readers (row display), so it stayed.

**Noted, acted on in the PR body**
- The picker loses key/tempo search. Deliberate, and now said out loud rather than left for a musician to discover.

Deferred items tracked: none — the plumbing question this PR declines to answer is #1440's own second half, and the reasoning is in the PR body; reopen or raise fresh if a third picker appears.

jonyardley:
CI's **Comment density** check failed at 0.23 against a 0.15 threshold, so I cut the comments to fit rather than leaning on the local bypass: 16 comment lines down to 9, on the same code. What survives is the reason the query scope clears rather than narrows, the `onAppear` capture hazard, and why the test's fixture words are disjoint. The check passes on its own now and the bypass note is out of the PR body.

jonyardley:
CI's iOS build failed: `LibraryItemView` gained a `photoId` field on main after this branch was cut, and the search test's hand-written fixture didn't have it. Merged main in and rebuilt that fixture by mutating `LibraryItemView.previewExercise` rather than adding another literal, so the next field added to the type costs no edit here. Gates re-run green on the merged tree, and the comment-density check passes without a bypass.

Issue #1447, opened 2026-08-29
Accented titles sort to the end of the library


jonyardley:
Sort the Library by title and a piece called "Étude" lands after "Waltz", after every plain-ASCII title in the library. A musician scrolling to E will not find their Études, Valses or Ländler.

`sort_library_items` (`crates/intrada-core/src/app.rs:1083`) compares `title.to_lowercase()` byte by byte, so any title starting with a non-ASCII letter sorts after every ASCII one. Piano repertoire has plenty of them.

The fix belongs in the core, not the shell, since the shell now mirrors the core's comparator deliberately (#1445). Whatever collation the core adopts, `sortedLikeTheLibrary` in `ios/Intrada/Core/LibraryItemView+Sort.swift` has to follow it, and its table test should gain an accented case.

Raised in the self-review of #1445.

jonyardley:
Claimed. Branch: `claude/library-title-collation-1447`, stacked on #1448 since it edits the same Swift comparator.

PR #1448, opened 2026-08-29
ios(picker): break sort ties the way the Library does (#1445)

### LINKED ISSUE #1445: Exercise picker sorts differently from the Library when keys tie
Sort by **Last practised** in the Library, then open **Choose one** on a piece and sort the **Add exercises** sheet the same way. Exercises that have never been practised come out in a different order in the two places.

The core's sort has a defined tiebreaker when the sort key is equal: newest first, then id (`sort_library_items`, `crates/intrada-core/src/app.rs:1102`). The picker filters and sorts its own candidates and its comparator has no tiebreaker at all (`isOrderedBefore`, `ios/Intrada/Views/Components/LinkedExercisePickerSheet.swift:196`), so ties land wherever Swift's sort leaves them.

Same class as the search divergence fixed in #1440, one function below it.

Fix: give the picker the core's tiebreaker, with a table test that ties resolve the same way in both. This does not need the pickers unified.

jonyardley:
Closes #1445.

## What changed

Sort by **Last practised** on the Library screen, then open **Choose one** on a piece and sort **Add exercises** the same way: exercises that have never been practised came out in one order in one place and another in the other.

The core resolves an equal sort key with a tiebreaker — newest first, then id (`app.rs:1102`). The picker sorts its own candidates and had no tiebreaker at all, so ties landed wherever Swift's sort left them. `sortedLikeTheLibrary` mirrors the core's comparator, tiebreaker included, and the picker uses it.

One deliberate consequence: title comparison follows the core's lowercased compare rather than the localised one the picker had, because the point is that the two screens agree. That makes the picker agree with the Library on something the Library gets wrong — accented titles sort after every ASCII one, so an Étude lands after a Waltz. That is a core-side fix, tracked as #1447 rather than papered over here.

Same class as the search divergence in #1440, one function below it. Nothing else in the picker reimplements a core rule.

## Verification

- New table test built around ties, with the fixture ordered against every assertion. Mutation-tested by deleting each branch in turn: removing either half of the tiebreaker fails the two tie tests, removing `.lowercased()` fails the title test, and flattening the last-practised comparison fails the never-practised test.
- Drove the sheet on the simulator to confirm sorting still works; the tie behaviour itself is what the unit tests are for, since the seeded data has no ties.

Coverage: no Rust changed. iOS is in Codecov's ignore list; the gate here is the table test above.

### DIFF
--- a/docs/status.md
+++ b/docs/status.md
@@ -39,6 +39,13 @@ audit backlog and its five-phase build order.
 
 ## Recently landed
 
+- #1445 — **the exercise picker no longer orders exercises differently from the
+  Library**. Sorting by **Last practised** put exercises that have never been
+  practised in one order on the Library screen and another in the **Add
+  exercises** sheet, because the sheet sorts its own candidates and had no rule
+  for what happens when the sort key ties. It now uses the core's: newest
+  first, then id. Same class as the search divergence in #1440, one function
+  below it.
 - #1440 — **search means the same thing in the exercise picker as in the
   Library**. Choosing an exercise for a piece (**Choose one** on the piece)
   opens a sheet whose search only read the exercise's title and its key/tempo
--- a/ios/Intrada/Core/LibraryItemView+Sort.swift
+++ b/ios/Intrada/Core/LibraryItemView+Sort.swift
@@ -1,0 +1,39 @@
+import Foundation
+import SharedTypes
+
+extension Array where Element == LibraryItemView {
+  /// The core's `sort_library_items` (`app.rs`), for a sheet that sorts its own
+  /// candidates: ties fall back to newest first, then id, so the same library
+  /// never comes out in a different order here than on the Library screen
+  /// (#1445).
+  func sortedLikeTheLibrary(by sort: LibrarySort) -> [LibraryItemView] {
+    sorted { a, b in
+      let primary = compareField(a, b, sort.field)
+      if primary != .orderedSame {
+        return sort.direction == .ascending
+          ? primary == .orderedAscending : primary == .orderedDescending
+      }
+      if a.createdAt != b.createdAt { return a.createdAt > b.createdAt }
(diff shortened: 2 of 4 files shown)


jonyardley:

**Verdict: ready with fixes** — both applied, second commit on the branch.

**What it confirmed by reading both comparators**
- The Swift matches `sort_library_items` on the subtle half: the tiebreaker applies to the *directed* result, so newest-first/id-ascending survive a descending primary sort instead of flipping with it.
- `?? ""` is right for the core's `Option` ordering — empty sorts before every RFC3339 timestamp the app can produce, which is what `None < Some(_)` gives.
- The two tie tests hold up under deletion, and `tiebreakIgnoresDirection` genuinely constrains the non-reversal.

**Fixed inline — two of the four tests were fake coverage**
- `titleOrdering` was named for case-insensitivity but its fixture ("scales" vs "Arpeggios") sorted the same way with `.lowercased()` deleted, so the one deliberate behaviour change in this PR was unguarded. Fixture is now "arpeggios" vs "Scales", which flips without it.
- `neverPractisedSortsEarliest` passed on the tiebreaker alone: its never-practised item was also the newer one, so deleting the whole last-practised comparison still gave the expected order. The never-practised item is now the older one.
- Re-mutated both after fixing, plus both halves of the tiebreaker. Each deletion fails the test that names it.

**Also**
- Trimmed the sort helper's doc comment and put the reason for dropping `localizedCaseInsensitiveCompare` at the call site rather than only in the commit message.

**Raised and tracked, not fixed here**
- The reviewer pointed out that this PR makes the picker agree with the Library on collation the Library gets wrong: `to_lowercase().cmp()` is byte order, so Études and Valses sort after every ASCII title. Correct intermediate state, wrong end state, and it belongs in the core. Opened as #1447.

Deferred items tracked: #1447

Issue #1446, opened 2026-08-29
Adding a photo while you are writing the piece, not after


jonyardley:
Phase A of [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md) puts the photo on the piece detail screen only. Add and Edit have no photo affordance, so a musician who has just typed a title has to save, tap into the piece, and add the photo there. Jon raised this while reviewing the phase A design (2026-08-29); this is spec open question 4, and the part of it phase A deliberately left alone.

## Why it was not done in phase A

Two reasons, and the second is the stronger one.

1. **It is a core change, not a screens change.** `ItemFormScaffold` collects field values and dispatches exactly one event on confirm (`Add` or `Update`). `SetPhoto` was deliberately kept out of `UpdateItem` in #1443, so a photo row in the form cannot ride the form's own submit. On Edit it would write before Save and Cancel would not undo it, breaking the sheet's contract that everything is provisional. On Add there is no item id to name yet, so the shell would have to sequence a write after the create resolves — a domain decision, in Swift.

2. **The shape inverts in phase B.** Once the page is read for title, composer and tempo, the photo is not a field you fill alongside them: it is the thing that fills them. The add surface's photo affordance is then an entry point *above* the form ("Scan a page" → recognition → the form comes back filled in), not a row inside it. Building the row now means deleting it in B.

## What to decide

- **a. Fold it into phase B**, as the confirm-sheet conversation spec open question 2 already flags. The scan entry lands on the add surface once there is something to scan *for*.
- **b. Do it now as its own change**, which means `Add` carrying an optional `photo_id` through the core (FFI + validation, domain-sensitivity override).

Leaning **a**. Adjacent to #1390 (one-pass create), which reworks the same screen.

jonyardley:
Unblocked in part. Spec open question 2 was answered by Jon on 2026-08-29 (full note on #1436): recognition **pre-fills the existing add form** rather than opening a confirm sheet of its own.

That settles the shape this issue was waiting on. Capture on the create surface is an **entry point above the form** — scan a page, then edit what it read — not a photo row sitting beside Title and Composer. Which is the reason this was deferred out of phase A in the first place: the row would have been the wrong thing, built and then deleted.

So option **a** from the description is now the live plan: this folds into phase B (#1436) rather than needing `Add` to carry an optional `photo_id` of its own.

PR #1449, opened 2026-08-29
ios(photo): keep the page you practise from, on the piece (#1355)

jonyardley:
Phase A **screens** of [`specs/piece-from-photo.md`](https://github.com/jonyardley/intrada/blob/main/specs/piece-from-photo.md), for #1355. Follows #1443, which merged while this was being built, so it targets main directly. **No recognition of any kind** — that is phases B to D.

Design first, as the spec asks: [the capture and display canvas](https://claude.ai/code/artifact/62d14f8d-b86c-4d87-9f49-4ab917bf15eb), seven frames drawn against the real `Theme.swift` values.

## What landed

- **The Photo card**, between the key/tempo rows and the chord chart. The two things you play from sit together, and the practice data stays above the fold. Same anatomy as the chord chart card next door, down to the `Add` / `Change` header button.
- **One menu for every route**: Scan the page (`VNDocumentCameraViewController`, which finds the page edges and flattens the perspective, so it is not just "Camera"), Choose a photo, and Remove once there is one. No separate "Replace" concept to learn.
- **Full screen on tap**, with pinch and double-tap zoom. A 244pt thumbnail of a page of music is not readable, so the viewer is the feature, not a nicety. New `viewerBackdrop` token — warm near-black from the ink family, because cream around a photograph tints how you read the page.
- **Remove asks first.** A scanned page was never saved to Photos, so removing it is unrecoverable for the user — the one case design-principles T6 keeps a confirm for. The message says what it costs ("You'll need to scan or choose it again") rather than the generic "This can't be undone".

## Bytes first, then the core

`PhotoCard.save` mints the ulid, writes the file, and only then sends `SetPhoto`. An item can therefore never end up naming a file that was never written. A failed write reports to Sentry, surfaces in the card, and stores nothing. A `SetPhoto` the core rejects keeps its error on screen instead of feeling like it landed (`store.send(_:onSuccess:)`, the surface-don't-swallow rule).

## Two spec open questions answered

- **6, HEIC or JPEG** → **JPEG, quality 0.8, long edge capped at 2048px.** "What the camera produces natively" does not apply: the scanner hands back a `UIImage` and the library route arbitrary bytes, so we re-encode either way. HEIC's size win is real but it has no encoder on some simulator configurations the tests run on, and JPEG is safer for a future export. A page lands at roughly 400 to 600 KB. Re-encoding also drops the source EXIF, so no photo carries the location it was taken at into the app container.
- **4, where the camera lives** → **the detail screen only, in this phase.** A photo row cannot ride `ItemFormScaffold`'s own submit (`SetPhoto` was kept out of `UpdateItem`, and on Add there is no item id yet), so it is a core change; and phase B inverts the shape anyway, since the photo stops being a field beside title and composer and becomes the thing that fills them. Deferred as #1446.

## Three departures from the spec's plan

- **The library route is SwiftUI's `PhotosPicker`, not a `PHPickerViewController` wrapper.** It runs out of process and needs no photo-library permission at all, so only `NSCameraUsageDescription` goes into `project.yml`. One fewer bespoke bridge, and better privacy. Verified on the simulator: the picker opens with no prompt; the scanner prompts with our own string, "To scan the page you practise from."
- **The shell mints the photo's ulid** (`ios/Intrada/Core/Ulid.swift`), because it needs the id before it can write the file — offline-first invariant 3, and what #1443's own description assumed. This is the first id the shell has ever minted, so it is pinned twice: against the ulid spec's published worked example, and through the **live bridge** against the core's `validate_photo_id`, which is the only reader that matters.
- **The bytes are written shell-side, not through a persistence `Effect`.** The spec's invariant-5 paragraph reads as though the write is core-driven; it is not, because the bytes must never cross the bridge and the shell needs the id first. The failure is still surfaced rather than faked — a failed write sends no event at all. Flagged because it matters for phase B, where `ReadPhoto { photo_id }` has the shell read a file the core assumes is there.

## Offline-first checklist

- [x] **1** No `Http` on this path — writes go through `SetPhoto`/`ClearPhoto`, which #1443 already asserts issue none.
- [x] **2** Nothing deletes a photo file: replacing, clearing and deleting the item all leave the bytes. Reaping is #1442.
- [x] **3** Client-minted ulid, minted in the shell, validated in the core.
- [x] **4** No merge or reconciliation logic added to Swift; the shell writes bytes and sends events.
- [x] **5** A failed file write surfaces and sends nothing — there is no path that fakes success.
- [x] **6** All new code, local-first only.

## Testing

- `UlidTests` — the spec's `01ARYZ6S41` worked example, the Crockford alphabet, time ordering, and uniqueness within one millisecond.
- `PhotoFileStoreTests` — write/read round-trip, the long-edge cap, and **the cap is measured in pixels, not points**. That last test found a real bug in my first version: `UIImage.size` is points, so a 3x image sailed past a cap it was nine times over. Fixed, and the test kills the regression.
- `testRealBridgeAcceptsAUlidTheShellMinted` — a Swift-minted id through `LiveBridge` into the core's validator. A stub bridge cannot catch this class of break (#846).
- `testPhotoCardStates` — component-level snapshot of both load-bearing states. Component rather than another whole-screen reference: the card is what changes, and those two states are the only thing that can regress independently.
- Eleven existing detail-screen references re-recorded, one at a time rather than by class, so history only gains the PNGs that actually changed. One of them (`testExerciseDetailWithStepsAccessibilitySize`) turned out to be order-dependent and had to be recorded in suite context — worth knowing if it bites again.

**Driven on the simulator, not just tested**: add via the picker, display, full screen, pinch zoom, Change menu, Remove with confirm, back to empty. Then **against real on-device data** (`SEED=0`) with a full app restart — the photo came back, which is the check that actually matters here.

Gates: `just check`, `just ios-fmt-check`, `just ios-test-full` — all green on the committed tree.

Coverage: `ios/` is Codecov-ignored and this PR is Swift-only, so expect no patch-coverage signal. The Swift tests above are the coverage. **Known gap**: `PhotoCard` itself has no unit coverage — the write-then-`SetPhoto` ordering, the reject path and the failure banner are exercised only by driving the simulator, which is why the run above walks each of them.

## Not in this PR

- Recognition of any kind (phases B to D).
- A photo while you create or edit a piece: #1446.
- Reaping orphaned files: #1442.
- Moving the JPEG encode off the main thread: #1450.
- A success and warning case for `SuccessFeedback`: #1451.

#1355 closes when this lands.


jonyardley:

Reviewed `e473719..cd7835e`. **No blockers.** The reviewer independently simulated the ULID bit arithmetic (`1469918176385` → `01ARYZ6S410000000000000000`, and `0xFFFFFFFFFFFF` → `7ZZZ…` confirming the 2-bit pad keeps the first character at or below `7`) and confirmed the `UInt32` accumulator never carries more than 12 significant bits, so there is no overflow. It also confirmed the points-vs-pixels handling and, usefully, traced the error path further than I had: `Store.process` resolves the persistence effect **synchronously**, so the `errorSeq` comparison in `send(_:onSuccess:)` genuinely catches a `PersistenceOutput::Failed` and not just a validation reject. Offline-first invariant 5 holds end to end.

Nine Important findings. **Eight fixed in 61bac1d**, one was a documentation gap.

### Fixed

1. **A real gesture bug.** `magnify.onEnded` reset `pan` to zero when zoom returned to 1 but left `dragStart` holding the old offset, so the next drag jumped by it. `toggleZoom` got this right; the pinch path did not. Both now go through one `settle()`. Reproduced and re-verified on the simulator: pinch in, pan, pinch out, pinch in, drag — the page now stays put.
2. **Pan was unbounded.** At zoom 1.1 you could flick the page off screen and be left staring at the backdrop, recoverable only by a double tap nothing advertises. Now clamped to the overhang the zoom actually creates. Verified with a hard flick.
3. **The Done button failed WCAG AA.** `viewerBackdrop.opacity(0.55)` over a white page composites to about `#818181`, which is **3.4:1** against `onAccent` — and the comment above it claimed the scrim existed precisely so nothing had to be assumed about the contrast beneath. Raised to 0.8, which is 8:1 in the same worst case. I recomputed this independently before accepting it; the reviewer's number was right.
4. **A disk read on every re-render.** `State(initialValue:)` evaluates its argument on *every* `init`, and `LibraryDetailScreen` re-creates `PhotoCard` on each view-model change and each of its eight `@State` toggles — so a `createDirectory` plus a half-megabyte file read ran on the main actor each time, and the result was thrown away. `PhotoFileStore.image` now caches. Nothing invalidates it, because every save mints a fresh ulid, so a written id is one nothing can already hold.
5. **A swallowed diagnostic.** `try? await item.loadTransferable(...)` discarded the error while both sibling paths reported to Sentry. Now `do`/`catch` with `report`.
6. **A dangling `photo_id` was unreachable.** After a restore from backup, or a future reaping bug, the id is in the core and the file is not: the card rendered "no photo yet" and hid Remove behind the very id it needed to clear. The header and Remove now key off the stored id, and the card says the photo is missing. This is the #846 shape and the best finding in the review.
7. **The design reference.** `viewerBackdrop` is now in `intrada-design-system.dc.html`, which the Design Workflow requires alongside a new token.
8. **Spec open question 5** ("who reaps orphan files — answer before phase A ships") was left open. Answered: a later explicit pass, tracked as #1442, and nothing reaps in phase A.

Plus three Nits taken: `url(for:)` refuses an id that is not a 26-character Crockford ulid (it becomes a path component, and a GRDB-hydrated id never passes the core's validator again — now covered by a traversal test); the clock is clamped so a device set before 1970 cannot trap the ulid constructor; and the store's tests clear up after themselves instead of leaving a JPEG per run.

### 9. An undocumented third departure

The spec's offline-first paragraph says "a failed file write resolves `Failed`, never a faked `Ack`", which reads as though the photo write is a core-driven persistence `Effect`. It is not: the bytes are written shell-side, because they must never cross the bridge and the shell needs the id before it can name a file. The failure is still surfaced rather than faked — a failed write sends no event at all. Recorded in the spec's phase A bullet, because it matters for phase B, where `ReadPhoto { photo_id }` has the shell read a file the core assumes is there.

The reviewer also judged the two documented departures, and agreed with both: `PhotosPicker` is a straight improvement, and shell-side ulid minting does not violate the dumb-pipe rule, since a ulid is an identifier format rather than a domain decision and `validate_photo_id` requires a ulid specifically as a traversal defence.

### Where I pushed back

- **"Is the Remove confirm justified at all?"** (T6 says confirm only the irreversible.) Keeping it. A `VNDocumentCameraViewController` scan is never written to Photos, so removing a scanned page really is unrecoverable for the user. The reviewer also read "You'll need to scan or choose it again" as naming the mechanism rather than the consequence; I read it the other way — the consequence of removing it *is* having to get it again, and it says more than the generic "This can't be undone" the item delete uses.
- **"The empty state is missing rule 7's what-to-do and its action button."** It mirrors `chartEmptyState` exactly — one purpose sentence, with the action as the header button. That card is the sibling directly below it on the same screen, and matching it is worth more here than satisfying the rule twice over in two different shapes.
- **"An empty Photo card is permanent chrome on every item."** Deliberate. It is the only affordance that tells you the feature exists, and the chord chart card next door is unconditional on the same screen for the same reason.

### Test quality

The reviewer mutation-tested each new assertion by **deleting** the line it guards, not inverting it: delete `* image.scale` and `measuresTheCapInPixelsNotPoints` fails; delete the `guard longest > longestEdge` and `leavesAPageThatIsAlreadySmallEnoughAlone` fails; delete the downscale and `bringsACameraSizedPageDownToTheLongEdgeCap` fails; any change to the pad width, byte order or alphabet fails `encodesTheSpecsWorkedExample`. No test here survives deletion of what it names.

The honest gap it named: **`PhotoCard` itself has no unit coverage** — the write-then-`SetPhoto` ordering, the reject path and the failure banner are exercised only by driving the simulator. That is the standard SwiftUI blind spot and I have not closed it; it is why the simulator run in the PR description walks every one of those paths, including a full app restart against real on-device data.

Gates re-run green after the fixes: `just check`, `just ios-fmt-check`, `just ios-test-full`, `scripts/check-snapshots.sh`.

Deferred items tracked: #1446, #1450, #1451

Issue #1454, opened 2026-08-29
A model-suggested field claims perfect confidence


jonyardley:
Phase B (#1436) sets `confidence: 1.0` on every field that came from `SuggestedFields` and survived the substring clamp (`crates/intrada-core/src/recognition.rs`, `clamped_text` / `clamped_tempo`).

That means a model suggestion can never trip `has_low_confidence`, so on the add form the strongest-looking claim is the one with the least evidence behind it — the opposite of what key decision 7 ("low confidence is shown, not hidden") is for.

It costs nothing today, because the shell always sends `suggested: nil` until phase C (#1437) wires Foundation Models. But it should be a deliberate decision in phase C rather than a default that arrived by omission.

The likely answer is to carry the OCR confidence of the lines the matched text spans — which needs the clamp to know which lines it matched, and it currently matches against all lines joined (see the spec's key decision 5, amended in #1436).

Blocks nothing. Do it as part of #1437.

PR #1455, opened 2026-08-29
core(photo): read a photographed page into title, composer and tempo (#1436)

jonyardley:
Phase B of [`specs/piece-from-photo.md`](../blob/main/specs/piece-from-photo.md), **core half**. The screens follow in a second PR. Closes nothing yet; #1436 stays open until the screens land.

## What this does

Photograph a page, and the core reads a title, a composer and a tempo out of it. The shell runs Vision and decides nothing; `read_fields` decides everything and is pure, so its tests need no device.

- **`RecognitionOperation::ReadPage { photo_id }`** resolving `RecognitionOutput` (`Page` / `Unsupported` / `Failed`). `Unsupported` is a normal outcome, not an error: the photo is still saved and the user types the fields.
- **`read_fields(&PageReading) -> PhotoDraft`**, geometry heuristics only. The largest text in the top band is the title; a `Music by` line is the composer; a tempo word or a number following an `=` is the tempo. Every field carries where it came from and how confident the read was.
- **The substring clamp** (key decision 5). A field the on-device model suggests, from phase C onward, is accepted only when it appears verbatim in the OCR text; otherwise it is discarded and the geometry heuristic stands. A model can choose, never invent, and that is structural rather than hoped for.
- **Vision in the shell** as a dumb pipe: recognised text with normalised geometry and confidence, origin flipped to top-left, nothing else.

Nothing recognised is written anywhere. The draft reaches the shell as a view projection; only the user pressing **Add** writes an item.

## Spec open questions answered

**2 — confirm sheet or pre-filled form?** Answered by Jon on the issue: recognition **pre-fills the add form**; no new surface. Folded into the spec here, ahead of the screens PR. This also unblocks #1446. What is left for the design conversation is narrower: how a recognised field is marked, whether a weak read is marked differently again, and whether the mark clears once the user edits it.

**3 — tempo from a note glyph?** Answered conservatively in the implementation: a bpm is read only when it follows an `=`. The glyph OCRs to junk or vanishes; the equals survives. A bare number near the top of a page is a page number or a bar number at least as often as a tempo.

## Departures from the pinned contract

Both are in the spec.

- `DraftField<T>` is split into `TextDraftField` / `TempoDraftField`: facet typegen emits no generics across the bridge.
- `ReadPhoto` takes only a `photo_id`, never an item id. On the add path there is no item to name yet, which follows from open question 2's answer.

## Against the offline-first checklist

- Recognition runs entirely on-device and issues no `Http` — test-enforced (`reading_a_page_issues_no_http`), invariant 1.
- No new persisted table or column, so invariants 2 and 3 are untouched; nothing is written at all on this path (`reading_a_page_never_writes_an_item`).
- Interpretation, the clamp and the state machine are all core, not shell — invariant 4.
- A photo the shell cannot read resolves `Failed`, never a page of no lines, which the core would read as a blank — invariant 5.
- All new code, local-first only — invariant 6.

## Tests

43 Rust tests in `recognition.rs`, four Swift tests driving real Vision over a page drawn in the test, one real-bridge round trip and one Store effect-loop test.

Every load-bearing line was mutation-tested by deleting it (not inverting it): the top-band filter, the credit filter, the substring clamp, the credit word-boundary check, the confidence damping, the tempo confidence minimum, the id correlation, the composer band check and the whole-line tempo rule each take a test down when removed. `bpm_in` was mutation-tested by substituting the naive "first number in the line" a future reader would plausibly write, which takes down three.

`every_read_field_is_one_the_create_form_accepts` is a table test against the consumer: seven pages a user would actually photograph, asserted against `validate_create_item`, not against the heuristics that produced them.

Coverage: expect high patch coverage on `recognition.rs`. Gaps are the Swift under `ios/` (Codecov-ignored) and the `Unsupported` arm, which no device can produce until phase C exists to be unavailable.

## Self-review

Posted as a comment below. Four findings were applied inline; the reviewer reproduced the first against the real app.

### DIFF
--- a/crates/intrada-core/src/app.rs
+++ b/crates/intrada-core/src/app.rs
@@ -125,9 +125,15 @@ pub enum Event {
     SessionStoreWritten(PersistenceOutput),
 
     // ── On-device recognition ────────────────────────────────────────
-    /// The shell read the page. `Unsupported` and `Failed` are outcomes the
-    /// confirm surface shows, not errors that mute the banner.
-    PhotoRead(RecognitionOutput),
+    /// The shell read the page. Carries the `photo_id` it was asked to read, so
+    /// a result can only ever land on the read that asked for it — two scans in
+    /// flight would otherwise show one page beside the other's fields.
+    /// `Unsupported` and `Failed` are outcomes the form shows, not errors that
+    /// mute the banner.
+    PhotoRead {
+        photo_id: String,
+        output: RecognitionOutput,
+    },
     /// The user finished with (or backed out of) the recognised draft.
     DiscardPhotoDraft,
 }
@@ -377,15 +383,18 @@ impl Intrada {
             },
 
             // ── On-device recognition ────────────────────────────────
-            Event::PhotoRead(output) => {
-                let crate::model::PhotoRecognition::Reading { photo_id } =
-                    std::mem::take(&mut model.photo_recognition)
-                else {
-                    // Nothing is waiting on this read: the user has already
-                    // moved on, and a late result must not repopulate a sheet
-                    // they have left.
+            Event::PhotoRead { photo_id, output } => {
+                // Anything but "still reading this exact photo" is a result
+                // nothing is waiting on — a scan the user backed out of, or one
+                // superseded by a later scan. Leave the state untouched: the
+                // read that *is* current must survive its predecessor landing.
+                if model.photo_recognition
+                    != (crate::model::PhotoRecognition::Reading {
+                        photo_id: photo_id.clone(),
+                    })
+                {
                     return crux_core::render::render();
-                };
+                }
 
                 model.photo_recognition = match output {
                     RecognitionOutput::Page(page) => crate::model::PhotoRecognition::Ready {
--- a/crates/intrada-core/src/app.rs
+++ b/crates/intrada-core/src/app.rs
@@ -25,10 +25,11 @@ use crate::http;
 use crate::model::{
     build_active_session_view, build_blocks, build_summary_view, entry_to_view, session_to_view,
     BuildingSetlistView, ItemPracticeSummary, LibraryItemView, LinkedExerciseView, Model,
-    PieceRefView, ScaffoldPreviewView, ScaffoldSpecView, SessionStatusView, SetSourceStatus,
-    ViewModel,
+    PhotoRecognitionView, PieceRefView, ScaffoldPreviewView, ScaffoldSpecView, SessionStatusView,
+    SetSourceStatus, ViewModel,
 };
 use crate::persistence::{self, PersistenceOperation, PersistenceOutput};
+use crate::recognition::{self, RecognitionOperation, RecognitionOutput};
 
 /// Root Crux application for the music practice library.
 #[derive(Default)]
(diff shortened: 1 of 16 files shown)


jonyardley:
## Self-review (code-reviewer subagent, `requesting-code-review`)

Reviewed `4efca08..09d18a2`, with comment-policy violations treated as Blockers. Verdict was **with fixes**: four issues, all now applied inline in `78bd68a`.

### Applied inline

**1 (Critical) — a read could land on a photo it did not come from.** `RecognitionOutput` carried no id, so the handler took whichever `Reading { photo_id }` was in the model. The reviewer reproduced it against the real app: with two scans in flight, photo B was shown beside fields read off photo A, and B's own result was then dropped. Compounding it, `std::mem::take` ran before the guard, so a late result destroyed a live read rather than being ignored. `Event::PhotoRead` now carries the `photo_id` it was asked to read and only lands when that read is still current, mutating nothing otherwise. New test `a_read_never_lands_on_a_photo_it_did_not_come_from`, mutation-verified.

**2 (Important) — a lyric starting with "By" became the composer.** Unlike the title and tempo heuristics, the composer scanned the whole page, and a bare `by` is ordinary English: "By the light of the silvery moon she waits" read as the composer at confidence 0.72, above the low-confidence threshold, so the form would not even have marked it as uncertain. A bare `by` now only counts in the top band; `Music by` and the other explicit credits still run the whole page.

**3 (Important) — a title containing a marking word lost the title and invented a tempo.** Matching a marking anywhere in a line meant "Swing Low, Sweet Chariot" read as tempo `Swing` with no title at all, and so did "Ballad of the Sad Young Men" and "Allegro Barbaro". A marking is now only read from a line that is a tempo instruction and nothing else, with a qualifier list so "Medium Swing" and "Lento ma non troppo" still read. Both pages are now in the consumer table test, which is where the reviewer rightly pointed out they should have been.

**4 (Important) — no real-bridge test for the new bridge types.** Nine new types cross the wire and `f32` is a new scalar on it. `testRealBridgeReadPhotoFillsTheDraft` now drives the whole round trip through `LiveBridge` and asserts the f32 confidence survives to the projection.

Minors also applied: a guessed bpm takes a stricter floor (20) than one the user typed, the image decode moved off the main actor alongside Vision, `chart_text` has a real bound instead of `usize::MAX`, and credit prefixes are matched without mapping a byte index between a string and its lowercase.

Also fixed a drive-by the reviewer spotted outside the range: phase A's `validate_photo_id` was inserted between `validate_piece_host`'s doc comment and its signature, leaving one function undocumented and the other carrying a stray paragraph about linked exercises.

### Confirmed clean

Comment policy (no blockers), the dumb-pipe boundary, `f32`'s bincode codec against the generated Serde runtime, path-traversal guards on both sides, and both documented departures from the pinned contract.

### Deferred

Model-suggested fields are hard-coded to `confidence: 1.0`, so a phase C suggestion could never be marked as a weak read. It costs nothing today, because the shell always sends `suggested: nil` until #1437 wires Foundation Models, but it should be phase C's deliberate decision rather than a default arrived at by omission.

Deferred items tracked: #1454

Issue #1456, opened 2026-08-29
A new iOS test file silently never runs


jonyardley:
Every `just` recipe that builds the iOS app runs `xcodegen generate --use-cache` (six sites in the `justfile`). The cache key is `project.yml`, and adding a Swift file does not change `project.yml` — so a **newly added test file is not in the target, and `just ios-test` passes without ever running it**.

It fails silently in the worst direction: the suite goes green, and the new tests look like coverage forever.

Hit while building #1436's screens. `ios/IntradaTests/ItemFormReadFieldTests.swift` was added, `just ios-test` went green, and a mutation test (deleting the line the tests were written to constrain) *stayed* green. Only `cd ios && xcodegen generate` without `--use-cache` put the file in the target, after which the same mutation failed two tests as it should.

The same trap applies to a new source file, though that usually announces itself as a compile error at the call site instead.

Worth fixing, roughly in order of preference:

1. Make the cache key include the file list under `ios/Intrada` and `ios/IntradaTests`, not just `project.yml`.
2. Drop `--use-cache` from the test recipes and keep it only where the inner-loop cost is worth it.
3. Failing both, a CI check that regenerates without the cache and fails if `Intrada.xcodeproj` changes.

Related: #1192 (the green-stamp skip) is a different mechanism but the same failure shape — a gate that thinks it has run something it has not.

jonyardley:
Working on branch `ios-test-target-1456`.

PR #1457, opened 2026-08-29
ios(photo): fill the add form from a photographed page (#1436)

### LINKED ISSUE #1436: Read a photographed page into title, composer and tempo (on-device OCR)

jonyardley:
Phase B of [`specs/piece-from-photo.md`](../blob/main/specs/piece-from-photo.md), **screens half**. Stacked on #1455 (the core); GitHub retargets this to `main` when that merges. Closes #1436 and unblocks #1446.

## What this does

**Scan a page** sits above the fields on the add screen. Photograph the page, the core reads it, and the form arrives holding the title, composer and tempo it could make out. Nothing is written until you press **Add**.

- Every field the photo filled says **From the photo** underneath it.
- A weak read wears the same mark, dimmed.
- The mark clears the moment you type in that field.
- A value you typed *before* scanning is never overwritten.

Design settled with Jon on 2026-08-29 against a mock of the real screen: [the canvas](https://claude.ai/code/artifact/578843ee-7f0a-4c56-8b16-d2e0cdf92fd3). Both options for the weak read were built and Jon picked the dimmed one.

## One thing worth your eye

The dim is `inkFaint`, which `Theme.swift` documents as failing WCAG AA (2.9:1) and reserves for eyebrow labels. That is a real cost, and it lands on the read the user most needs to catch.

It was raised before the choice was made and the choice stands. What this PR adds rather than changes: a weak field also says so to VoiceOver — *"Read from the photo, but not clearly. Worth checking"* — so the signal is never contrast alone. If the dim proves too quiet in use, the fix is the mark's wording, not the colour.

## Where the scan entry sits

Above the form, not as a row inside it. That is spec open question 4's answer playing out: once a page is read it is not a field beside Title and Composer, it is what fills them. #1446 was waiting on exactly this.

## Consolidation

`PhotoCard` (the photo on a piece, from phase A) already owned the capture path — the document scanner, the library picker, and writing the bytes before anything names them. Rather than the second surface copying the first, that half moved to `PhotoCaptureSources` and both use it. `ItemFormScaffold` gained a header slot with a defaulted `EmptyView`, so the edit screen is untouched.

## Tests

- **Six unit tests** on the provenance rules (`ItemFormReadFieldTests`), each mutation-tested by deleting the line it constrains: removing the clear-on-edit write-through and removing the don't-overwrite guard each take tests down.
- **Two new snapshots**: the form as it arrives from a read page (composer weak), and the same after correcting the composer. Three existing Add screen references re-recorded for the scan entry.

Coverage: no new Rust. `ios/` is Codecov-ignored, so expect no patch-coverage signal on this PR.

## A gate that was not running

While mutation-testing the above, deleting the line the new tests were written to constrain left the suite **green**. The cause was not a weak test: every `just` recipe runs `xcodegen generate --use-cache`, the cache keys on `project.yml`, and a newly added Swift file therefore never entered the target. `just ios-test` passed without running the new file at all.

Regenerating without the cache put it in, after which the same mutation failed two tests as it should. Filed as #1456 — it fails silently in the worst direction, and it is not specific to this PR.

## Self-review

Posted as a comment below.


jonyardley:
## Self-review (code-reviewer subagent, `requesting-code-review`)

Verdict was **no, with fixes**: two of the three controls on the new surface did not do what they said. Both are fixed inline, along with a third the reviewer rated Important.

### Applied inline

**1 (Critical) — the scanner had no way out of the piece detail screen.** Extracting the capture path out of `PhotoCard` dropped `scanning = false`; `VNDocumentCameraViewController` does not dismiss itself, so after Save *or* Cancel the user was stuck in a full-screen scanner. `ScanPageEntry` had the line, `PhotoCard` did not. **No gate could have caught it**: the scanner is unavailable on the simulator, so it is invisible to snapshots and XCUITests and ships broken to devices only. Rather than adding the line back at both sites, both presentations now live in a `.photoCapture` modifier that owns the dismissal — the call site cannot forget it again. **This one still wants a device check before merge**, since nothing automated reaches it.

**2 (Critical) — Rescan could not change anything the first scan filled.** Guarding each write on `isEmpty` cannot tell a value the user typed from one the previous scan wrote; `readFrom` is exactly the record that can. Scan page A, realise it was wrong, Rescan page B: the thumbnail swapped and the fields did not. Now a field is written when empty **or** when it still holds an earlier read. Two tests, both mutation-verified.

**3 (Important) — `Failed` and `Unsupported` had no surface.** Both rendered as "Scanned page", so a read that never happened looked identical to one that worked, after the user had waited through a spinner. Every outcome the core models now says something, plus the empty-`Ready` case (a photograph of something that is not a page reads cleanly and finds nothing). `Unsupported` also drops the Change trigger — another page reads no better on a device that cannot read one.

Also applied: the mark is spoken as the **field's own hint**, not only as a sibling element a rotor user would never land on — which is the whole of what compensates for a contrast-only signal, so it mattered more than its Important rating; the refill is keyed on the projection rather than draft equality, so re-picking the same library file is not a silent no-op; the draft is discarded from the sheet's `onDismiss` rather than the add screen's `onDisappear`, which the scanner's own full-screen cover can also trigger; and the trigger reads "Change", matching `PhotoCard` and naming both things its menu does.

**`has_low_confidence` removed.** The reviewer spotted that this PR was the consumer it was added for and went per-field instead, leaving it with a test as its only reader. It came in with the core PR, which has since merged, so it is removed here rather than left sitting in main.

The reviewer also gave `loadImage` its missing reader: `testScanPageEntryStates` snapshots all six states of the entry, which is where the outcome copy above is actually held.

### Confirmed clean

The dumb-pipe boundary (no Swift thresholds on confidence — the core sends `weak`), the no-silent-write guarantee traced through every mutation path, and the `@Observable` write-through under both the computed getter and `@Bindable`'s bindings.

### Merge state

Main moved under this branch when #1455 merged. Cleared by merging main in, not rebasing (`docs/status.md` is union-merged, so the flag was the usual false one and resolved with no manual work). Now `MERGEABLE`.

### Not changed

**Em dashes in comments.** CLAUDE.md bans them; `ios/Intrada` has ~192 already, and the new files match their neighbours rather than being the only ones that differ. Flagging rather than silently choosing either way — worth a sweep or a rule change, not a divergent style in one PR.

Deferred items tracked: #1456 (a new iOS test file silently never runs — found here when a mutation test stayed green), #1458 (the dimmed mark is below AA contrast), #1459 (SF Symbols do not scale with Dynamic Type)

